PR #394, opened 2015-01-04
work around with xdg-open behaviour under Linux

Chilledheart:
unlike mac's open program, xdg-open won't fork a new subprocess to open
the file will block until the application returns, so we won't wait for it
and we need another approach to check if it works

### DIFF
--- a/src/utils/utils.cpp
+++ b/src/utils/utils.cpp
@@ -42,6 +42,50 @@ const char *kCcnetConfDir = "ccnet";
 const char *kCcnetConfDir = ".ccnet";
 #endif
 
+#ifdef Q_WS_X11
+/// \brief call xdg-mime to find out the mime filetype X11 recognizes it as
+/// xdg-mime's usage:
+/// xdg-mime query filetype <filename>
+/// stdout: mime-type
+bool getMimeTypeFromXdgUtils(const QString &filepath, QString *mime)
+{
+    QProcess subprocess;
+    QStringList args("query");
+    args.push_back("filetype");
+    args.push_back(filepath);
+    subprocess.start(QLatin1String("xdg-mime"), args);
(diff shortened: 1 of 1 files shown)


src/utils/utils.cpp
```diff
@@ -59,16 +103,34 @@ bool openInNativeExtension(const QString &path) {
...
+    // the file will block until the application returns, so we won't wait for it
+    // and we need another approach to check if it works
+
+    // find out if the file can be opened by xdg-open, xdg-mime
+    // usually they are installed in xdg-utils installed by default
+    QString mime_type;
+    if (!getMimeTypeFromXdgUtils(path, &mime_type))
+        return false;
+    // don't open this type of files from xdg-mime
+    if (mime_type == "application/octet-stream")
```
lins05: On one ubuntu system I see this:

```sh
$ xdg-mime query filetype ~/.bashrc
text/plain; charset=us-ascii
```
Please handle this case as well.

Chilledheart: Ops, that's bad. But I think it is rare and due to broken mime database from desktop environment. Cleaning `~/.local/share` might help.

Chilledheart: I find out some interesting facts about `xdg-utils` from the comments of [chrome-wrapper](https://github.com/Chilledheart/chromium/blob/master/chrome/tools/build/linux/chrome-wrapper#L76) file from chromium project which has strong support for linux :

> We include some xdg utilities next to the binary, and we want to prefer them
> over the system versions when we know the system versions are very old.
 
From the results of searching `xdg-utils`-related topics, `xdg-mime` even might return with an empty output. And if it fails, it will fall back to `showInGraphicalShell`` .

Does this information help you?

lins05: That's fair enough.

PR #396, opened 2015-01-04
Add sort functionality

Chilledheart:
include #395 , so it should rebase before being merged.

### DIFF
--- a/qt.css
+++ b/qt.css
@@ -451,6 +451,16 @@ FileTableView QHeaderView::section {
     font-size: 12px;
 }
 
+FileTableView QHeaderView::down-arrow {
+    width: 15px;
+    height: 5px;
+    subcontrol-position: center right;
+}
+
+FileTableView QHeaderView::up-arrow {
+    subcontrol-position: center right;
+}
+
 FileTableView::item {
     border-top: 1px solid #e0e0e0;
 }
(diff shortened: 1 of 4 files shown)


lins05:
After deleting a file it becomes like this:

screen shot 2015-01-04 at 17 52 57

lins05:
And seafile-applet crashes after selecting two files and deleting them together.

Chilledheart:
it's a bug with deleting. should be fixed now.

Chilledheart:
It is what the dialogs looks like on os x:

screen shot 2015-01-06 at 10 04 35

lins05:
It still has the problem I mentioned. This is what looks like after deleting a file.

screen shot 2015-01-06 at 10 46 50

After deleting two files together:

screen shot 2015-01-06 at 10 48 39

Chilledheart:
Sorry, but it is fine here.

lins05:
I can reproduce the problem under both windows and mac. Can you check it on another system?

Chilledheart:
Thanks for the lack of the details of the bug and I had to spend hours thinking about what's going wrong. So I can repeat the error by these steps:
- select some items to be deleted
- click sort indicator
- (don't deselect and) right-click these items and choose remove
 
Thanks to the qt, this bug gets fixed after using the methods such as `beginRemoveRow`. BTW I have improved the related code.

Again, nice catch for the glitching bug because it would cost me lots of debugging time if the bug left uncovered.

Chilledheart:
I find out it still has some bugs.

Chilledheart:
It should be fixed now.

Issue #372, opened 2014-12-22
seafile-client 4.0.4-message translation


lavoisier16:
After an user action, seafile-client displays a message on desktop screen.
A part of these messages are not translated in french translation ( words  Added , Modified, Deleted .... ) see screenshot

![screenshot- 22122014 - 09 18 24](https://cloud.githubusercontent.com/assets/7565826/5522683/67b46416-89bd-11e4-991c-275b9ea45ebd.png)

Issue #270, opened 2014-11-05
French translation


lavoisier16:
Hi
In seafile client settings ( advanced tab )  french translation chains are truncated because windows is too narrow.
Please use the last translation from transifex
regards

PR #398, opened 2015-01-04
add a helper script to update translations

Chilledheart:


scripts/update-i18n.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+function regenerate_source() {
+    SEAFILE_PROJECT="seafile-client.pro"
+    rm -f $SEAFILE_PROJECT
+    qmake -project -o $SEAFILE_PROJECT
+    lrelease $SEAFILE_PROJECT $QUIET_FLAGS
+    lupdate $SEAFILE_PROJECT $QUIET_FLAGS
+    rm -f $SEAFILE_PROJECT
+}
```
lins05: You can use "make update-ts". That's what I do each time I need to update translations. See https://github.com/haiwen/seafile-client/blob/v4.0.5/CMakeLists.txt#L354

Chilledheart: If you write some script functionality inside CMake before, you will find it is horrible and hard to maintain.
Besides previous update-ts target failed for many translations update, e.g. https://github.com/haiwen/seafile/issues/847

Chilledheart: Or notably, this one #270

Chilledheart: Yes. notably this one #372 .

In this case, I had to use `lrelease` to make it work and fixed the problem.

lins05: Regarding to #372, how did that fail? It has been working for a very long time (since the beginning when we have i18n support in seafile-applet), and I don't think it could fail without a good reason.

Chilledheart: I think `lupdate` requires the qmake project file in some degree. and I believe whether the qmake project file exists or not makes difference. And in fact, many of sources of translations hadn't updated since months ago.

Sadly, I didn't look into the issue any further since the solution (regenerate the pro file and execute lrelease and lupdate) solves this issue. So I wrote this script not only for the automatizing the translations-updating progress, but for recording this solution we found (sometime we might forget how to solve the issues we faced before, we haven't to need to solve them again because we can everything from this script)

lins05: It seems to me [this change](https://github.com/haiwen/seafile-client/commit/f18789950425a852e905b3cb531f880d87ef914b#diff-af3b638bc2a3e6c650974192a53c7291L340) has something to do with #372 . 

"update-ts" doesn't work properly in current master, but with the above change reverted, "update-ts" works without any problem.

Chilledheart: Nope, it had broken for at least two months.

Chilledheart: I can confirm the bug appeared from v4.0.1.

Chilledheart: #401 should clean up with #372 

the script is added with "update-ts" support

PR #400, opened 2015-01-05
fix transfer rate to zero issue

### LINKED ISSUE #358: Sync speed not shown correctly on "hover" of tray icon
see image:

shows for instance 0 KB/s instead of 2100 KB/s. In the bottom of the clinet it is shown correctly. And the German translation is also not the best. Better would be `Lädt $bib hoch`.
Client version: 4.0.4

 Not an important issue but still a bug.

Chilledheart:
Closes #358.

### DIFF
--- a/src/message-listener.cpp
+++ b/src/message-listener.cpp
@@ -40,22 +40,6 @@ static int parse_seafile_notification (char *msg, char **type, char **body)
     return 0;
 }
 
-static bool
-collect_transfer_info (QString *msg, char *info, char *repo_name)
-{
-    char *p;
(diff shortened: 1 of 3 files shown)


lins05:
We used to show "Library A uploading, X KB/s, Library B downloading, Y KB/s" in the tray icon tooltip. After this change it would only show a general "Uploading X KB/s, Downloading Y KB/s". Anyway that's ok with me.

Chilledheart:
Yeah, I know the difference but at least the more generic solution works as expected.

PR #404, opened 2015-01-06
add language option

Chilledheart:
screen shot 2015-01-06 at 16 51 18

Alongside some fixes are applied:
- fix TR_tr and zh_TW translation

### DIFF
--- a/src/i18n.cpp
+++ b/src/i18n.cpp
@@ -1,0 +1,145 @@
+#include "i18n.h"
+#include <QTranslator>
+#include <QLibraryInfo>
+#include <QApplication>
+#include <QSettings>
+#include <QDebug>
+
+namespace {
+const char* langs[] = {
+    NULL, //reserved for system locale
+    "ca",
+    "de_DE",
+    "en",
+    "es",
+    "es_AR",
+    "es_MX",
+    "fr_FR",
+    "he_IL",
+    "hu_HU",
+    "is",
+    "it",
+    "ko_KR",
+    "nl_BE",
+    "pl_PL",
+    "pt_BR",
+    "pt_PT",
+    "ru",
+    "sk_SK",
+    "uk",
+    "zh_CN",
+    "zh_TW",
+    "tr",
+    NULL
+    };
+void saveCurrentLanguage(int langIndex) {
+    QSettings settings;
+
+    settings.beginGroup("Language");
+    settings.setValue("current", QString(langs[langIndex]));
+    settings.endGroup();
+}
+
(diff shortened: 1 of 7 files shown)


src/i18n.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+int I18NHelper::preferredLanguage() {
+    return loadCurrentLanguage();
+}
+
+void I18NHelper::setPreferredLanguage(int langIndex) {
+    const QList<QLocale> &locales = getInstalledLocales();
+    if (langIndex < 0 || langIndex >= locales.size())
+        return;
+    saveCurrentLanguage(langIndex);
```
lins05: I'd rather save the language string, instead of the integer index. The index if relative and may change in the future.

ui/settings-dialog.ui
```diff
@@ -192,6 +192,41 @@
...
+      <attribute name="title">
+       <string>Language</string>
+      </attribute>
+      <layout class="QVBoxLayout" name="languageTabLayout">
+       <item>
+        <layout class="QHBoxLayout" name="languageLayout">
+         <item>
+          <widget class="QLabel" name="mLanguageLabel">
+           <property name="text">
+            <string>Language (need restart)</string>
```
lins05: Is it possible to avoid the restart after the user changes to another language?

src/i18n.cpp
Chilledheart: No, it is saved as a string in QSettings.And the int number is only used at runtime and calculated dynamically from the string.

see more at [saveCurrentLanguage](https://github.com/haiwen/seafile-client/pull/404/files#diff-f0f193787a82c9b195cd700c580219aeR35)

ui/settings-dialog.ui
Chilledheart: It is hard to implement because it requires to add event filters for all the widgets.

src/i18n.cpp
lins05: Sorry, I didn't read that carefully.

lins05:
The code looks good to me. But I'm thinking does it worth adding a new "International" tab, which contains only one settings. @freeplant What do you think?

Chilledheart:
In fact there is no good place in other tabs to place it (it will break the consistency of the ui look).  Or do you have any more practical advice?

src/i18n.cpp
Chilledheart: No worry about it.

Chilledheart:
Closed in 4444ebe

PR #407, opened 2015-01-07
fix more memory leaks

Chilledheart:


Chilledheart:
This should fix the last definite memory leaks reported by valgrind. The others are possible memory leaks.

PR #410, opened 2015-01-08
[feature] exit gracefully when pressing ctrl+c

Chilledheart:
qt won't register these handlers for us. and when it receives SIGINT,
it exits directly without destroying and releasing objects and memory.

btw, under linux, qt uses glib as its internal event dispatcher, so
calling signal (2) won't work.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -21,6 +21,46 @@
 #include "application.h"
 #endif
 
+#ifndef Q_OS_WIN32
+#include <signal.h>
+#endif
+
+namespace {
+
+#ifndef Q_OS_WIN32 // windows don't need these handlers
+
+void sigintHandler(int /* unused*/)
+{
+    qWarning("[signal] SIGINT received, exiting");
+    QCoreApplication::exit();
+}
+
+bool setSignalHandler()
+{
+    struct sigaction sigint_action;
+    sigint_action.sa_handler = sigintHandler;
+    sigemptyset(&sigint_action.sa_mask);
+    sigint_action.sa_flags = 0;
+    sigint_action.sa_flags |= SA_RESTART;
+
+    // 0 if okay, -1 if failed
+    if (sigaction(SIGINT, &sigint_action, 0) < 0)
+        return false;
+    return true;
+}
+
+#else // Q_OS_WIN32
+
+bool setSignalHandler()
+{
+    return true;
+}
+
+#endif // Q_OS_WIN32
+
+} // anonymous namespace
+
+
 #define APPNAME "seafile-applet"
 
 namespace {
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -22,6 +22,44 @@
...
+bool setSignalHandler()
+{
+    struct sigaction sigint_action;
+    sigint_action.sa_handler = sigintHandler;
+    sigemptyset(&sigint_action.sa_mask);
+    sigint_action.sa_flags = 0;
+    sigint_action.sa_flags |= SA_RESTART;
+    if (sigaction(SIGINT, &sigint_action, 0) != 0) // 0 if okay, -1 if failed
+        return false;
+    return true;
```
lins05: Can we simplify these into `signal(SIGINT, sigintHanlder)`?

lins05: > btw, under linux, qt uses glib as its internal event dispatcher, so
> calling signal (2) won't work.

I've tried that under ubuntu 14.04, and it seems to work fine.

Chilledheart: nice, i tried it on os x 10.10 and fedora 21 and it worked as expected.

Chilledheart: In fact there is a document from qt for it : https://qt-project.org/doc/qt-4.8/unix-signals.html

freeplant:
“exit without destroying and releasing objects and memory”: Why this is matter? After program exit, the memory will be released by the operating system.

Chilledheart:
No, (when sigint is triggerred) it is calling `_exit` which is abnormal exit instead of standard `exit`

Chilledheart:
Here is a small example from [stackoverflow] which you can save to a file `test.cc` and compile it by `g++ -o test test.cc`:
```cpp
#include <iostream>
#include <signal.h>
#include <cstring>
#include <unistd.h>

bool quit = false;    // signal flag

void got_signal(int)
{
    quit = true;
}

class Foo
{
public:
    ~Foo() { std::cout << "destructor\n"; }
};

int main(void)
{
    struct sigaction sa;
    memset( &sa, 0, sizeof(sa) );
    sa.sa_handler = got_signal;
    sigfillset(&sa.sa_mask);
    sigaction(SIGINT,&sa,NULL);

    Foo foo;    // needs destruction before exit
    while (true)
    {
        // do real work here...
        sleep(1);
        if( quit ) break;    // exit normally after SIGINT
    }
    return 0;
}
```

The program calls method `sigaction` to register `SIGINT` signal handler as this PR does. when you press `CTRL+C`, it should print `destructor`, which means the dtor of the c++ object is destroyed normally.

Howevery, if you comment out the `sigaction`, and recompile it, you will find it fails to print anything. It is an abnormal exit now.

see more: 
- [origin post from stackoverflow][stackoverflow]
- http://www.cons.org/cracauer/sigint.html

[stackoverflow]: http://stackoverflow.com/questions/4250013/is-destructor-called-if-sigint-or-sigstp-issued

PR #411, opened 2015-01-08
fix a bundle of memory leaks

Chilledheart:
notably the QMovie one which leaks lots of memory


src/ui/repo-tree-model.cpp
```diff
@@ -61,8 +61,11 @@ RepoTreeModel::RepoTreeModel(QObject *parent)
 }
 RepoTreeModel::~RepoTreeModel()
 {
-    if (item(kIndexOfVirtualReposCategory) != virtual_repos_category_)
-        delete virtual_repos_category_;
+    // if we miss it, just add it and
+    // let dtor of repotreemodel to delete it automatically
+    if (item(kIndexOfVirtualReposCategory) != virtual_repos_category_) {
+        insertRow(kIndexOfVirtualReposCategory, virtual_repos_category_);
+    }
```
lins05: Doing this(inserting a row) in the destructor really feels counter-intuitive to me.

src/ui/starred-files-tab.h
```diff
@@ -48,7 +49,7 @@ private slots:
     QWidget *loading_failed_view_;
     QWidget *empty_view_;
 
-    GetStarredFilesRequest *get_starred_files_req_;
+    QScopedPointer<GetStarredFilesRequest> get_starred_files_req_;
```
lins05: I see `QScopedPointer` is used in multiple places to avoid memory leak. Can we make add a bool field to SeafileApiRequest, and make it delete itself after the request is finished, if that field is true?

Chilledheart: Ehh, `QScopedPointer` (aka something like `std::unique_ptr`) is simple and useful to keep the lifetime of the object allocated from heap, because you know how it is managed and when it is destroyed (it is destroyed only the wrapper object is destroyed). The way you mentioned (deleting itself after completed) is regaining the memory faster. However, in the later way, it is much more complicate that you need to add additional slot to delete itself but you can hardly tell if others slots(or handlers) require the access of this object. So it can be tricky to make it memory-safe and correct.

In fact, I am good to use more code for more efficient stuff and if you want to do it, `deleterLater` instead of `operator delete` might help. If you want to keep in a simple way, it is good enough now..

src/ui/repo-tree-model.cpp
Chilledheart: When a object is being destroyed:

1. the code of its destructor(or dtor, in short)'s function body is executed
2. (after 1) the code at the end of function, i.e. `}`, is executed. it does destruction of all elements of the object.

So, in this case, the inserting code is executed before the the real part of dtor is executed.

Feel free if you have any questions

Chilledheart: I have reverted this part since I am not sure if the solution provided solves the memory leaks. If we come back again, make sure we eliminate this leak in the next patch.

PR #414, opened 2015-01-09
fix resync a repo when using http sync

lins05:
We need to pass `server_url` in the `more_info` field of the rpc call.


Chilledheart:
LGTM

PR #420, opened 2015-01-15
ui: no more toggle_main_window

### LINKED ISSUE #417: "Show main window" shouldn't change to "Hide main window"
This is a UI enhancement proposal. (For OS X, may be other OS concerned)
The tray icon has a "Show main window" entry when the window is hidden, which allow to display the window on top. So the entry menu change to "Hide main window".
But on OS X at least, when we switch to another app without closing the window, the entry menu remain "Hide main window". At this moment, to re-show the main window, we have to click to "hide main window", then click on tray, and click on "Show main window".
I suggest to have a single menu entry "Show main window" which just display the window on top and don't change to "Hide main window". If we want to hide the window we can just close it as it is possible to do actually.

Chilledheart:
Closes #417.

### DIFF
--- a/src/ui/tray-icon.cpp
+++ b/src/ui/tray-icon.cpp
@@ -118,8 +118,8 @@ void SeafileTrayIcon::createActions()
     quit_action_ = new QAction(tr("&Quit"), this);
     connect(quit_action_, SIGNAL(triggered()), this, SLOT(quitSeafile()));
 
-    toggle_main_window_action_ = new QAction(tr("Show main window"), this);
-    connect(toggle_main_window_action_, SIGNAL(triggered()), this, SLOT(toggleMainWindow()));
+    show_main_window_action_ = new QAction(tr("Show main window"), this);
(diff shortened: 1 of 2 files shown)


src/ui/tray-icon.cpp
```diff
@@ -352,13 +346,11 @@ QIcon SeafileTrayIcon::stateToIcon(TrayState state)
 #endif
 }
 
-void SeafileTrayIcon::toggleMainWindow()
+void SeafileTrayIcon::showMainWindow()
 {
     MainWindow *main_win = seafApplet->mainWindow();
     if (!main_win->isVisible()) {
```
lins05: This if condition should be removed, otherwise seafile applet won't be brought to front when the focus is on anthor application. E.g. start seafile applet -> click somewhere else -> click "show main window" => nothing happens

Chilledheart: Fine. I thought it was a qt's method. If it is a qt's method instead of a home-brewed method, I'd rather to do some check here.

Done in 89fee0c

PR #422, opened 2015-01-16
show the owner in repo_tree_view

Chilledheart:
Screen shot for private shares:

screen shot 2015-01-16 at 12 48 53

Screen shot for group shares:

screen shot 2015-01-16 at 12 50 43

### DIFF
--- a/src/ui/repo-item-delegate.cpp
+++ b/src/ui/repo-item-delegate.cpp
@@ -13,6 +13,7 @@
 #include "repo-tree-model.h"
 #include "rpc/rpc-client.h"
 #include "rpc/local-repo.h"
+#include "account-mgr.h"
 
 #include "repo-item-delegate.h"
 
@@ -51,6 +52,7 @@ const char *kTimestampColor = "#959595";
 const char *kTimestampColorHighlighted = "#9D9B9A";
 const int kRepoNameFontSize = 14;
 const int kTimestampFontSize = 12;
+const int kOwnerFontSize = 12;
 
 const char *kRepoItemBackgroundColor = "white";
 const char *kRepoItemBackgroundColorHighlighted = "#F9E0C7";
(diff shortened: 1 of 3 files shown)


src/ui/repo-item-delegate.cpp
```diff
@@ -218,6 +222,27 @@ void RepoItemDelegate::paintRepoItem(QPainter *painter,
                       &repo_desc_rect);
     painter->restore();
 
+    // Paint repo sharing owner for private share
+    // TODO: show only for private share
+    if (item->repo().owner != seafApplet->accountManager()->accounts().front().username) {
```
lins05: `accountManager()->currentAccount()` can help here.

lins05:
We should only display the name part the email address,  for example only show "foo" for "foo@bar.com"

src/ui/repo-item-delegate.cpp
Chilledheart: I know what, but `currentAccount` returns a new (and unnecessary) copy of Account while using `accounts().front()` doesn't (only reference)

Chilledheart:
Well, I am pretty con to that. It looks like no difference between group owner and person owner in that way.

Chilledheart:
Updated, no more group owner shown. cc @freeplant 

screen shot 2015-01-19 at 15 15 56

Chilledheart:
The owner should be shown only in the category "private share" now.

src/ui/repo-item-delegate.cpp
```diff
@@ -218,6 +222,27 @@ void RepoItemDelegate::paintRepoItem(QPainter *painter,
                       &repo_desc_rect);
     painter->restore();
 
+    // Paint repo sharing owner for private share
+    if (static_cast<RepoCategoryItem*>(item->parent())->categoryIndex() ==
```
lins05: We can use `item->repo().isShareRepo()` here.

Chilledheart: Yeah, but we want to show the owner precisely in the private shared category, not outside.

lins05: That's fair enough.

PR #423, opened 2015-01-16
fix the rare case of segment fault while uploading files

Chilledheart:
caused by the dialog automatically closing before the task is finished due to
the fact that the value is large than the maximum.

this patch adds guard to prevent it. and it also applies some human fixes for
the incorrect total_bytes during uploading (aka. not counting the network
 or http protocol cost).


Chilledheart:
Close for 4.0 branch

Chilledheart:
replaced by #425

PR #425, opened 2015-01-16
fix the rare case of segment fault while uploading files

Chilledheart:
caused by the dialog automatically closing before the task is finished due to
the fact that the value is large than the maximum.

this patch adds guard to prevent it. and it also applies some human fixes for
the incorrect total_bytes during uploading (aka. not counting the network
 or http protocol cost).

### DIFF
--- a/src/filebrowser/progress-dialog.cpp
+++ b/src/filebrowser/progress-dialog.cpp
@@ -44,6 +44,10 @@ FileBrowserProgressDialog::FileBrowserProgressDialog(FileNetworkTask *task, QWid
     initTaskInfo();
 }
 
+FileBrowserProgressDialog::~FileBrowserProgressDialog()
+{
+}
+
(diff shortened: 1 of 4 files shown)


lins05:
I still got the crash even after applying this patch. 

I did some debug and found  `onProgressUpdate` of FileBrowserProgressDialog is called after `onTaskFinished` has already been called.

Chilledheart:
Yeah. It means the task is cancelled when the value is larger than the maximum and the windows closes automatically after this, before the uploading is completed.

I have uploaded a new version, you can try with it.

Chilledheart:
It is broken. I will try to fix it.

Chilledheart:
use timer's timeout to filter out the progress events, it should work now

NOTE:

there is a bug with qt4's event loop. if you process the events too quickly, the event loop sender the events to the receiver even the receiver is no longer there. qt5 has no issue with this.

PR #426, opened 2015-01-16
close all FileBrowserDialogs when we are deleting the account

Chilledheart:


lins05:
branch name :+1: 

BTW, this PR reminds me of another issue: When I open the file browser in one account, copy an item, and switch to another account, and open a file browser again, the "copy" operation is still visible.

We should clear "files_to_be_pasted" when the user switches to another account.

Chilledheart:
I think it is a bug with the menu. While we allow FileBrowserDialogs from different accounts exist at the same time, clearing `files_to_be_pasted` might be not so good.

Chilledheart:
I think you mean `paste_action`.

BTW `paste_action` is visible but disabled in that case. In fact, it is intended to be visible, which tells the user that you have something in the clipboard but you are not able to paste them here.

Chilledheart:
So I don't think it is a actual bug. It is a different tone from UI designs.

PR #428, opened 2015-01-19
add support for restarting application by itself

Chilledheart:
<del>In deed, it is a quite large change.

Notably, the code of singletons should be two-checked.

cc @lins05 ,  should we put it onto branch master instead and do more tests before we can merge it?
</del>

A more convenient solution provided

### DIFF
--- a/src/seafile-applet.cpp
+++ b/src/seafile-applet.cpp
@@ -279,6 +279,20 @@ void SeafileApplet::errorAndExit(const QString& error)
     QCoreApplication::exit(1);
 }
 
+void SeafileApplet::restartApp()
+{
+    if (in_exit_ || QCoreApplication::closingDown()) {
+        return;
+    }
+
+    in_exit_ = true;
+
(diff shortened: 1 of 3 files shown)


src/seafile-applet.cpp
```diff
@@ -279,6 +279,20 @@ void SeafileApplet::errorAndExit(const QString& error)
...
+    }
+
+    in_exit_ = true;
+
+    QStringList args = QApplication::arguments();
+    args.removeFirst();
+    QProcess::startDetached(QApplication::applicationFilePath(), args);
+    QCoreApplication::quit();
+}
+
```
lins05: Currently when seafile-applet starts up, it would check if these is already another "seafile-applet" process running, and if so, tells the user "seafile is already running", and then exits. Is it a possible problem that the old process has not exited yet when the newly started process starts up?

Chilledheart: You can try it out. In fact, it just works :-P.

PR #435, opened 2015-01-22
qt5: build under gnu++11 mode

Chilledheart:
fix mingw build issue


Chilledheart:
- Fix mac build as well
- the binary libraries shipped with qt5.4.0 are linked against libc++ already. Linking programs with both of libstdc++ and libc++ might cause serious issues.

PR #437, opened 2015-01-24
qt5: fix conflicting included headers

Chilledheart:
fix conflicting included headers if we have qt5 and qt4 both in the path.

Notably this fixes the conflicting issues on the buildbot (where we are
using QT5's binary distribution while qt4-mac installed). After this we
don't need to deactivate port qt4-mac every time we do a qt5 release.

BTW macports's `port deactivate` is really slow and time consuming.


lins05:
I'm ok with this PR, but why can it change fix the problem?

Chilledheart:
Compilers are looking after headers by the order of including `-I`.

For example on macports:
we have OPENSSL_INCLUDE_DIRS at /opt/local/include
              QT4's headers ... at /opt/local/include
              QT5 ... at any other directory (outside macports)

When we are building against qt5, we don't want the compiler to use the qt4's headers.
But the compiler will scan the `/opt/local/include` first since in the cmake file it specifies
 `OPENSSL_INCLUDE_DIRS` first and then qt's include dirs. And if the compiler find the
file which has exactly the same name, it will think it is done and ignore the other candidates.
At the end, we are compiling qt5 program with (partial of) qt4's headers.

It looks like stupid, but it is the actual way the compile plays with. If we will want to change the rule,
 just tell cmake to include the qt's specific header dir before the global ones. This will solve the issue.

lins05:
Cool, thanks for the explanation.

lins05:
Sorry for the wrong click... Reopen and merged.

Chilledheart:
You are welcome.

PR #439, opened 2015-01-26
[api] fix a regression of renaming or deleting a filename contains plus sign(+)

Chilledheart:
Fix these issues:
- cannot rename a file which name contains "+"
- cannot remove a file which name contains "+"

Reasons:

- In Qt5, `QUrlQuery never encodes the space character to "+" and will never decode "+" to a space character`. You need do it manually.
- see more: http://doc.qt.io/qt-5/qurlquery.html

### DIFF
--- a/src/api/api-request.cpp
+++ b/src/api/api-request.cpp
@@ -45,15 +45,16 @@ void SeafileApiRequest::send()
 
 #if (QT_VERSION >= QT_VERSION_CHECK(5, 0, 0))
     if (!params_.isEmpty()) {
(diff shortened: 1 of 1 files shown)


lins05:
When I rename a file from "3+4.md" to "5+6".md in the file browser, it get renamed to "5 6.md" on the server.

Chilledheart:
And any more details excluding reporting the issue?

Chilledheart:
@lins05 Sorry, but I missed to handle with the form data. It should work as expected.

PR #450, opened 2015-02-07
ask the user whether to kill the previous seafile instance when found running

Chilledheart:


Chilledheart:
screen shot 2015-02-07 at 16 29 14
screen shot 2015-02-07 at 16 29 22

lins05:
We'd better call [do_stop()](https://github.com/haiwen/seafile-client/tree/3f4224c158b2cf3803dfb8195881fa2de8f24b80/src/utils/uninstall-helpers.cpp#L118) to politely asks the existing process to quite before kill it directly. This method would send a ccnet "quit" message which would cause the other seafile-applet process to [quit itself](https://github.com/haiwen/seafile-client/tree/3f4224c158b2cf3803dfb8195881fa2de8f24b80/src/message-listener.cpp#L122) when receiving it. 

If we find the existing process does not quit itself, the we can kill it.

Chilledheart:
Well, it has more business to do here which is not planed.

- any necessary calls that should be called before `do_stop`?
- what about the ui interaction? more stopping dialogs for error warnings?

lins05:
Call `do_stop`, wait for one second, if the existing process is still, kill it.

Chilledheart:
Okay, I will do it soon.

Chilledheart:
cc @lins05 I made some improvements. However I am not familiar with method `do_stop`, is it good enough now?

PR #451, opened 2015-02-09
remove unused images

Chilledheart:


lins05:
I'd rather have the previous python script to. Besides, we don't to need to use the external "ag" command, `git grep` is enough.

Chilledheart:
Python is really hard to write and debug. So I prefer to golang, which has more strong static checks but still compiles fast. 

If you don't want to use the tool, ignore it simply.

freeplant:
It is better to not introduce another language into the project. This is not a good for project maintenance.

Chilledheart:
@freeplant Okay, but I won't tend to write python scripts any more, because I don't want to waste lots of time writing and debuging on it. Instead I will set up separate repository to keep my golang scripts.

PR #460, opened 2015-02-14
refactor http requests query params handling

lins05:
This PR removes some duplicating of sending api requests when handling the difference of qt4 and qt5.


Chilledheart:
LGTM

BTW Qt5 uses `QList<QPair<QString, QString> >` to keep query items, and has several builtin methods to pass the value directly, e.g. via `void QUrlQuery::​setQueryItems(const QList<QPair<QString, QString> > & query)`. Is there any benefit we can get from dropping the Qt's `QList<QPair<QString, QString> >` for `QHash<QString, QString>`?

lins05:
Using a map instead of list of pairs may make it simpler when we want to find the param value of a give key, but actually I'm ok with both.

PR #461, opened 2015-02-14
use separate api token when using shibboleth

lins05:


lins05:
related: https://github.com/haiwen/seahub/pull/378

PR #467, opened 2015-02-26
Revert "work around with QTBUG-43073"

Chilledheart:
it is fixed in Qt 5.4.1

This reverts commit 05bd779925bbe277bf68146880cdf9fe55f31ddc.


lins05:
May be we should keep it for those still using 5.4.0 (using proper `#if` macro)? Just like what we do in https://github.com/haiwen/seafile-client/pull/468 .

Chilledheart:
I'd better not because it is only affected in Qt5.4.0 and, what's more, this workaround is totally not working at all.

PR #468, opened 2015-02-26
drop workaround with QTBUG-40449 if using qt 5.4.1 or later

Chilledheart:
- [QTBUG 40449](https://bugreports.qt.io/browse/QTBUG-40449) is fixed in qt5.4.1
- TODO remove this workaround and related code totally once qt 5.4.1 is widely used


lins05:
LGTM

PR #469, opened 2015-02-26
fix a bug with handwritten getParentPath

Chilledheart:


src/filebrowser/tasks.cpp
```diff
@@ -676,7 +676,7 @@ void PostFilesTask::startNext()
     QString file_name = QFileInfo(file_path).fileName();
     QString relative_path;
     if (use_relative_)
-        relative_path = file_path.left(file_path.size() - file_name.size());
+        relative_path = ::getParentPath(file_path);
```
lins05: We also have similar code [here](https://github.com/haiwen/seafile-client/tree/4aa47d3c9bca9e1d919a32799949909e44ddef3a/src/filebrowser/tasks.cpp#L270), should it be changed too?

Chilledheart: updated

PR #475, opened 2015-03-03
Sync a subfolder into new library

Chilledheart:
- <del>still in progress</del>

Edited: 

- finished

### DIFF
--- a/src/account-mgr.cpp
+++ b/src/account-mgr.cpp
@@ -58,6 +58,11 @@ int AccountManager::start()
         return -1;
     }
 
+    // enabling foreign keys, it must be done manually from each connection
+    // and this feature is only supported from sqlite 3.6.19
+    sql = "PRAGMA foreign_keys=ON;";
+    sqlite_query_exec (db, sql);
+
     sql = "CREATE TABLE IF NOT EXISTS Accounts (url VARCHAR(24), "
         "username VARCHAR(15), token VARCHAR(40), lastVisited INTEGER, "
         "PRIMARY KEY(url, username))";
--- a/src/api/requests.cpp
+++ b/src/api/requests.cpp
@@ -23,6 +23,8 @@ const char *kApiPingUrl = "api2/ping/";
 const char *kApiLoginUrl = "api2/auth-token/";
 const char *kListReposUrl = "api2/repos/";
 const char *kCreateRepoUrl = "api2/repos/";
+const char *kGetRepoUrl = "api2/repos/%1/";
+const char *kCreateSubrepoUrl ="api2/repos/%1/dir/sub_repo/";
 const char *kUnseenMessagesUrl = "api2/unseen_messages/";
 const char *kDefaultRepoUrl = "api2/default-repo/";
 const char *kStarredFilesUrl = "api2/starredfiles/";
@@ -191,17 +193,44 @@ void DownloadRepoRequest::requestSuccess(QNetworkReply& reply)
 }
 
 /**
+ * GetRepoRequest
+ */
+GetRepoRequest::GetRepoRequest(const Account& account, const QString &repoid)
+    : SeafileApiRequest (account.getAbsoluteUrl(QString(kGetRepoUrl).arg(repoid)),
+                         SeafileApiRequest::METHOD_GET, account.token)
+      , repoid_(repoid)
+{
+}
+
+void GetRepoRequest::requestSuccess(QNetworkReply& reply)
+{
+    json_error_t error;
+    json_t *root = parseJSON(reply, &error);
+    if (!root) {
+        qWarning("failed to parse json:%s\n", error.text);
+        emit failed(ApiError::fromJsonError());
+        return;
+    }
+
+    QScopedPointer<json_t, JsonPointerCustomDeleter> json(root);
+    QMap<QString, QVariant> dict = mapFromJSON(json.data(), &error);
+    ServerRepo repo = ServerRepo::fromJSON(root, &error);
+
+    emit success(repo);
+}
+
+/**
  * CreateRepoRequest
  */
-CreateRepoRequest::CreateRepoRequest(const Account& account, QString &name, QString &desc, QString &passwd)
+CreateRepoRequest::CreateRepoRequest(const Account& account, const QString &name, const QString &desc, const QString &passwd)
     : SeafileApiRequest (account.getAbsoluteUrl(kCreateRepoUrl),
                          SeafileApiRequest::METHOD_POST, account.token)
 {
-    this->setFormParam(QString("name"), name);
-    this->setFormParam(QString("desc"), desc);
+    setFormParam(QString("name"), name);
+    setFormParam(QString("desc"), desc);
     if (!passwd.isNull()) {
         qWarning("Encrypted repo");
-        this->setFormParam(QString("passwd"), passwd);
+        setFormParam(QString("passwd"), passwd);
     }
 }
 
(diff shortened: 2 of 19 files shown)


src/repo-service.cpp
```diff
@@ -42,16 +42,19 @@ void RepoService::stop()
...
     }
 
-    const std::vector<Account>& accounts = seafApplet->accountManager()->accounts();
-    if (accounts.empty()) {
-        in_refresh_ = false;
-        return;
+    if (!account) {
+        const std::vector<Account>& accounts = seafApplet->accountManager()->accounts();
+        if (accounts.empty()) {
+            in_refresh_ = false;
```
lins05: This above code can be simplified to `accountManager->currentAccount()->isValid()`.

src/api/requests.cpp
```diff
@@ -220,6 +248,37 @@ void CreateRepoRequest::requestSuccess(QNetworkReply& reply)
...
 
 /**
+ * CreateSubrepoRequest
+ */
+CreateSubrepoRequest::CreateSubrepoRequest(const Account& account, const QString &name, const QString &repoid , const QString &path, const QString &passwd)
+    : SeafileApiRequest (account.getAbsoluteUrl(QString(kCreateSubrepoUrl).arg(repoid)),
+                         SeafileApiRequest::METHOD_GET, account.token)
+{
+    this->setUrlParam(QString("p"), path.endsWith('/') ? path : path + "/");
+    this->setUrlParam(QString("name"), name);
```
lins05: Unnecessary `this` pointer.

src/repo-service.cpp
Chilledheart: Yeah, but there is still a unnecessary copy when calling `currentAccount`. Why not `hasAccount`?

Chilledheart: I still don't think it is more convenient to call `currentAccount` here.

src/api/requests.cpp
Chilledheart: Updated in the latest patches.

src/account-mgr.cpp
```diff
@@ -119,6 +120,8 @@ int AccountManager::saveAccount(const Account& account)
 
 int AccountManager::removeAccount(const Account& account)
 {
+    RepoService::instance()->onAccountRemoval(account);
+
```
lins05: What about adding a new signal "accountRemoved" and let `ReposService` listen to this signal?

src/account-mgr.cpp
```diff
@@ -172,6 +175,7 @@ bool AccountManager::setCurrentAccount(const Account& account)
 
 int AccountManager::replaceAccount(const Account& old_account, const Account& new_account)
 {
+    RepoService::instance()->onAccountChange(old_account, new_account);
```
lins05: Same as above, I think using a signal is better for enforcing modularity.

src/filebrowser/file-browser-dialog.cpp
```diff
@@ -962,7 +977,38 @@ void FileBrowserDialog::onDirentsMoveSuccess()
...
+    data_mgr_->createSubrepo(folder_name, repo_.id, ::pathJoin(current_path_, folder_name), QString());
+}
+
+void FileBrowserDialog::onCreateSubrepoSuccess(const ServerRepo &repo)
+{
+    LocalRepo r;
+
+    // if we have not synced before
+    bool has_local = false;
+    if (seafApplet->rpcClient()->getLocalRepo(repo.id, &r) == 0 && r.isValid()) {
```
lins05: Can be simplified by `hasLocalRepo(repo.id)`.

src/repo-service.cpp
```diff
@@ -11,29 +14,134 @@
...
+    {
+        repo_id_ = rhs.repo_id_;
+        parent_repo_id_ = rhs.parent_repo_id_;
+        parent_path_ = rhs.parent_path_;
+        return *this;
+    }
+
+    const QString& repo_id() const { return repo_id_; }
+    const QString& parent_repo_id() const { return parent_repo_id_; }
+    const QString& parent_path() const { return parent_path_; }
```
lins05: Please use `camelCase` instead of `under_score` for member function names.

src/repo-service.cpp
```diff
@@ -138,3 +295,132 @@ void RepoService::openLocalFile(const QString& repo_id,
...
+}
+
+void RepoService::startGetRequestFor(const QString &repo_id)
+{
+    bool was_empty = get_repo_reqs_.empty();
+    GetRepoRequest *req = new GetRepoRequest(seafApplet->accountManager()->currentAccount(), repo_id);
+    get_repo_reqs_.push_back(req);
+    connect(req, SIGNAL(success(const ServerRepo&)), this, SLOT(onGetRequestSuccess(const ServerRepo&)));
+    connect(req, SIGNAL(failed(const ApiError&)), this, SLOT(onGetRequestFailed(const ApiError&)));
+    if (was_empty)
```
lins05: `get_repo_reqs_.length() == 1` is enough, no need to use a `was_empty` variable.

src/repo-service.cpp
```diff
@@ -138,3 +295,132 @@ void RepoService::openLocalFile(const QString& repo_id,
...
+        server_repos_.push_back(fixed_repo);
+    }
+
+    delete req;
+    get_repo_reqs_.pop_front();
+
+    if (!get_repo_reqs_.empty())
+        get_repo_reqs_.front()->send();
+    else
+        emit refreshSuccess(server_repos_);
```
lins05: The code between line (315, 321) is the same as (330, 336), why not refactor them to a common subroutine?

src/repo-service.cpp
```diff
@@ -138,3 +295,132 @@ void RepoService::openLocalFile(const QString& repo_id,
...
+        return;
+    }
+
+    QString sql = "REPLACE INTO SyncedSubfolder(repo_id, parent_repo_id, url, username, parent_path) VALUES ('%1', '%2', '%3', '%4', '%5')";
+    sql = sql.arg(subfolder.id)
+             .arg(subfolder.parent_repo_id)
+             .arg(account->serverUrl.toEncoded().data())
+             .arg(account->username)
+             .arg(subfolder.parent_path);
+    sqlite_query_exec (synced_subfolder_db_, sql.toUtf8().data());
```
lins05: The path name may contain single quote `'` in it, so it's not safe to use `QString::arg()` to build the sql statement.  We could use `sqlite3_mprintf`.

https://www.sqlite.org/c3ref/mprintf.html

Chilledheart: Sure, I see the similar usage in `src/account-mgr.cpp`. Mayby we should change it as well, in a separate PR?

src/repo-service.cpp
Chilledheart: No, they are totally different.

src/repo-service.cpp
Chilledheart: `get_repo_reqs_` is a `std::list` aka double-link list. If calling `std::list::length` is avoidable,  using it is not so smart a option that it will iterates all its notes.

src/repo-service.cpp
Chilledheart: Okay, this class is used for repo-service.cpp internally and can be changed any time.

src/filebrowser/file-browser-dialog.cpp
Chilledheart: Sure.

src/account-mgr.cpp
Chilledheart: I did it for code simplicity when I wrote it. I saw the demands for the simplicity days ago. Have they changed yet?

src/repo-service.cpp
lins05: I think readable code is more important than the trivial performance gain. 

When I read this function, I see the `was_empty` variable is defined at the very beginning of the function, but used at the very end of the function, which is quite counter-intuitive, and makes me think:  "Why did he do this?" and takes me while to understand it.

Of course I'm not saying performance is not important, but considering today's cpu frequency, I really don't think we need to save cpu cycles everywhere for a desktop application at the cost of code readability.

Chilledheart: Assuming you know all of properties of double-link list, I think the code present is very readable. Of course, there is many common practice in C++ and it is one of them.

src/account-mgr.cpp
Chilledheart: Use foreign keys to avoid manually handling on these update/delete changes as recommended.

Chilledheart:
Rebased

lins05:
LGTM

PR #476, opened 2015-03-03
support logout the current account

lins05:
After logout, local repos would still be there, just no longer syncing with the server until the user re-logs into  the same account again.

Related:

- haiwen/seahub#408
- haiwen/seafile#1112


src/account-mgr.cpp
```diff
@@ -207,3 +231,36 @@ Account AccountManager::getAccountBySignature(const QString& account_sig) const
 
     return Account();
 }
+
+bool AccountManager::clearAccountToken(const Account& account)
+{
+    int i = 0;
```
Chilledheart: What about using `size_t` or `unsigned` here?

`std::vector<T>::size()` returns a `size_t` here, which is unsigned.

src/account-mgr.cpp
```diff
@@ -207,3 +231,36 @@ Account AccountManager::getAccountBySignature(const QString& account_sig) const
...
+            && accounts_[i].username == account.username) {
+            accounts_.erase(accounts_.begin() + i);
+            break;
+        }
+    }
+
+    Account new_account = account;
+    new_account.token = "";
+
+    accounts_.insert(accounts_.end(), new_account);
```
Chilledheart: Do you mean `accounts_.push_back(new_account);` here?

Chilledheart:
This code is fine with me. And I make some comments in the code if you want to improve a bit better.

PR #478, opened 2015-03-04
implement server-info support

Chilledheart:


src/account-mgr.cpp
```diff
@@ -148,17 +148,14 @@ bool AccountManager::setCurrentAccount(const Account& account)
 
 int AccountManager::replaceAccount(const Account& old_account, const Account& new_account)
 {
-    int i = 0;
-    for (i = 0; i < accounts_.size(); i++) {
-        if (accounts_[i].serverUrl.toString() == old_account.serverUrl.toString()
-            && accounts_[i].username == old_account.username) {
-            accounts_.erase(accounts_.begin() + i);
+    for (size_t i = 0; i < accounts_.size(); i++) {
+        if (accounts_[i] == old_account) {
```
lins05: Should not use `==` here. The `operator==` also compares the token, which we should not do here.

lins05: Never mind. It's ok to use `==` here.

src/ui/account-view.cpp
```diff
@@ -97,7 +97,7 @@ void AccountView::updateAccountInfoDisplay()
                                "\"color:#A4A4A4; text-decoration: none;\" "
                                "href=\"%1\">%2</a>").arg(href).arg(host);
 
-        mServerAddr->setText(text);
+        mServerAddr->setText(account.isPro() ? tr("%1 <small>pro version</small>").arg(text) : text);
```
lins05: Maybe `QString("%1 <small>%2</small").arg(text).arg(tr("pro version"))` is more suitable for translation?

src/account-mgr.cpp
```diff
@@ -207,3 +204,41 @@ Account AccountManager::getAccountBySignature(const QString& account_sig) const
...
+
+void AccountManager::serverInfoSuccess(const Account &account, const ServerInfo &info)
+{
+    for (size_t i = 0; i < accounts_.size(); i++) {
+        if (accounts_[i] == account) {
+            if (i == 0)
+                emit beforeAccountChanged();
+            accounts_[i].serverInfo = info;
+            if (i == 0)
+                emit accountsChanged();
```
lins05: The `accountsChanged` signal is meant to be sent when the user switches to another account, and emitting it here  would affect too many other components unnecessarily. We could have a new signal "accountInfoChanged", and let accountView or cloudView to connect to this new signal.

src/ui/account-view.cpp
Chilledheart: Yeah, it is great.

src/account-mgr.cpp
Chilledheart: Certainly, triggering `accountsChanged` signal again will require a lot of updates on all components. But if we use extra signals, it will involve a lot of complexities when you need to detect whether it has pro feature, and add extra code there. I cannot see it is great for programming and code simplicity. In short, although triggering `accountsChanged` is costly, it is worthy and I prefer to using it.

lins05: If you think it's too complex, then propose a simpler solution, that's ok. My bottom line is: doing unnecessary operations that can and should be avoided is unacceptable.

Chilledheart: Sorry, but I cannot catch you here.

lins05: As I mentioned, emitting `accountChanged` signal here would cause unnecessary operations on other modules which connects to this signal. For example, the avatar module would clear the in-memory cache of avatars, and the seahub notification module would send a request immediately to query unread notifications. 

To avoid these, my suggestion is: add another signal `accountInfoUpdated`, and emit this signal instead of `accountChanged`. account-view.cpp should listen to this new signal, so that the other modules would not be affected. The `accountChanged` signal is designed to be used when the user switches to another account.

> But if we use extra signals, it will involve a lot of complexities when you need to detect whether it has pro feature, and add extra code there.

I see your consideration here. But I don't think there are many places in the code where we need to write these possible extra code.

Chilledheart: I think we have discussed this topic before and made a decision there. I still don't know why you change your mind and what is so important that make you think using signal `accountInfoUpdated` instead is more correct.

Yes, but I am con to this change which involves new complexities and potential bugs. Anyway, if you insist to do it, I suggest you to do it in a separate PR as kind of improvement. According to your words, it shouldn't cost too much code there, so it is a pretty easy job. Above all be careful that don't introduce new bugs there.

PR #480, opened 2015-03-05
fix a bug that settings dialog still write discarded changes on settings

Chilledheart:
reproduce procedures:
- change language or proxy settings
- click cancel to discard current changes
- exit program directly
- get an alert that language or proxy settings has been changed and app requires a restart

This bug exists in previous versions as well.

### DIFF
--- a/src/ui/settings-dialog.cpp
+++ b/src/ui/settings-dialog.cpp
@@ -79,14 +79,32 @@ void SettingsDialog::closeEvent(QCloseEvent *event)
     this->hide();
 }
 
+void SettingsDialog::reject()
+{
+    // reset settings
+    // this fix a bug writing discarded settings when exiting program
+    readSettings();
+
+    QDialog::reject();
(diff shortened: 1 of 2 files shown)


lins05:
This PR should work, but I guess delete [this line](https://github.com/haiwen/seafile-client/tree/e34f3fa9f6e96714cdadafe6987dfdcd2b61153c/src/ui/settings-dialog.cpp#L76) is enough?

Chilledheart:
@lins05 I don't know this history of these code before and why they are added and what they fix. This patch is only to provide the fix the current bug without changing too much. At the first glance the solution should work, but I cannot guarantee it if the `guess` will work without breaking previous fixes. Anyway if you want more clean code as well as the correct logic, I think rewriting it or applying a better solution in the development branch is better.

Chilledheart:
It should be fixed in the incoming commit, closing.

PR #481, opened 2015-03-06
handle http proxy disconnection more gracefully

Chilledheart:
- handle unknown proxy error
- don't need a restart for proxy setting changes (gui part)

### DIFF
--- a/src/network-mgr.h
+++ b/src/network-mgr.h
@@ -1,0 +1,52 @@
+#ifndef SEAFILE_CLIENT_NETWORK_MANAGER_H
+#define SEAFILE_CLIENT_NETWORK_MANAGER_H
+#include <QObject>
+#include <vector>
+#include <QNetworkReply>
+class QNetworkAccessManager;
+class QNetworkProxy;
+class NetworkManager : public QObject {
+  Q_OBJECT
+public:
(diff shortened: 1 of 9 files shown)


src/network-mgr.h
```diff
@@ -0,0 +1,52 @@
...
+
+    // retry only once
+    bool shouldRetry(const QNetworkReply::NetworkError error) {
+        if ((error == QNetworkReply::ProxyConnectionClosedError ||
+             error == QNetworkReply::ProxyConnectionRefusedError ||
+             error == QNetworkReply::ProxyNotFoundError ||
+             error == QNetworkReply::ProxyTimeoutError ||
+             error == QNetworkReply::UnknownProxyError) &&
+            should_retry_) {
+            NetworkManager::instance()->reapplyProxy();
```
lins05: Why do we need reapply the proxy again here?

lins05:
We should not resend the request when it fails due to a proxy problem, because the server may have already handled the api request at that moment. e.g. when sending a request to create a library, resending the request it may cause two libraries get created.  We could just let the request fail, and let the user try the operation again.

Chilledheart:
We have a discuss internally, you can join it there.

PR #483, opened 2015-03-10
filter activities only for synced or owned repositories

Chilledheart:


src/ui/activities-tab.cpp
```diff
@@ -79,7 +82,21 @@ void ActivitiesTab::refreshEvents(const std::vector<SeafEvent>& events,
...
+    // TODO use a option to control this filter
+    std::vector<SeafEvent> filtered_events;
+    LocalRepo r;
+    ServerRepo repo;
+    for (size_t i = 0; i < events.size(); ++i) {
+        const QString &repo_id = events[i].repo_id;
+        repo = RepoService::instance()->getRepo(repo_id);
+        if ((repo.isValid() && repo.owner == seafApplet->accountManager()->currentAccount().username) ||
+            // TODO use RepoService::instace()->getLocalRepo to avoid RPC call
+            (seafApplet->rpcClient()->getLocalRepo(repo_id, &r) == 0 && r.isValid())) {
```
lins05: You can use `SeafileRpcClient::hasLocalRepo(repo_id)` here.

Chilledheart: done

lins05:
The code looks good to me. But what if all the events are filtered out, the events tab would show nothing, right? I think we can request more events from the server until we have at enough items.

Chilledheart:
@lins05 Well, it is more complicate to support with `load more events` at client side but not impossible. One more question, why those events are not filtered by the server side while we just need the filtered ones only? It seems more reasonable to filter events at the server side anyway.

Chilledheart:
It seems I get the task misunderstood, just closing it.

@lins05 just let me know if we are supported to add `load more events`. It looks like great.

PR #486, opened 2015-03-12
filter recent updated repos

Chilledheart:


src/ui/repo-tree-model.cpp
```diff
@@ -117,15 +118,23 @@ void RepoTreeModel::setRepos(const std::vector<ServerRepo>& repos)
...
 
-    QList<ServerRepo> list = map.values();
+    const QList<ServerRepo> list = map.values();
+    // use pointer to avoid additional object copys
+    QList<const ServerRepo*> filtered_list;
+    const QString owner = seafApplet->accountManager()->currentAccount().username;
+    Q_FOREACH(const ServerRepo& repo, list)
+    {
+        if (repo.owner == owner || seafApplet->rpcClient()->hasLocalRepo(repo.id))
+            filtered_list.push_back(&repo);
```
lins05: So if the user has no owned or synced repos, the "recently updated" category would be empty?

I think a better way is:

1. sort the repos in a way so that owned/synced repos are sorted before the others. And within owned/synced repos, sort them by last update time
2. list the first `kMaxRecentUpdatedRepos` of the above sorted repos under the "recently updated` category

This would achieve the goal, but would also work even if the user has no owned/synced repos.

Chilledheart: Yeah, but it was discussed with @freeplant minutes before.

PR #489, opened 2015-03-12
strip incorrect sRGB profile png files

Chilledheart:


lins05:
How do you strip it? I tried using the [strip-png.sh](https://github.com/haiwen/seafile-client/tree/tmp/scripts/strip-png.sh) script and it changes 81 files.

Chilledheart:
@lins05 they always changes if do it on striped files again.

Chilledheart:
@lins05 obviously, prog `convert` changes all png files no matter whether they are stripped or not..

PR #493, opened 2015-03-19
fix windows build under mingw32

Chilledheart:
`_WIN32_WINNT` is not set under mingw32. set its value to match Windows XP.

### DIFF
--- a/src/utils/process-win.cpp
+++ b/src/utils/process-win.cpp
@@ -1,3 +1,6 @@
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0501
+#endif
 #include <windows.h>
 #include <psapi.h>
 


lins05:
What does this fix?

Chilledheart:
@lins05 Yes it fixes the undefined function `GetProcessImageFileName` when compiling with gcc 4.4 and mingw32. This function is not available until we set `_WIN32_WINNT` to match Windows XP but mingw32 doesn't set it at all.

The similar cases are available at the [seafile codebase](https://github.com/haiwen/seafile).
- https://github.com/haiwen/seafile/blob/7de9c1888127c500181bf511b3b585482a83f5c5/daemon/client-migrate.c#L4
- https://github.com/haiwen/seafile/blob/9c76fc921c6c2eaa59c458b275aeba9a409b1979/common/obj-backend-fs.c#L1
- https://github.com/haiwen/seafile/blob/4ac9eb1b3a9fc54e976635c7fd3beddaf8787001/common/block-backend-fs.c#L3

BTW gcc 4.9 with mingw64 set `_WIN32_WINNT` for us. So it works even without this patch.

Chilledheart:
@lins05 more document is at https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745%28v=vs.85%29.aspx
screen shot 2015-03-19 at 17 03 49

PR #499, opened 2015-03-27
avatar: add support for auto update 

Chilledheart:
- fix a bug with copy assignment with Account

### DIFF
--- a/src/avatar-service.cpp
+++ b/src/avatar-service.cpp
@@ -11,14 +11,19 @@
 #include "utils/paint-utils.h"
 #include "utils/utils.h"
 
+#include <sqlite3.h>
 #include "avatar-service.h"
 
-namespace {
+static const int kCheckPendingInterval = 1000; // 1s
+static const char *kAvatarsDirName = "avatars";
+static bool loadTimeStampCB(sqlite3_stmt *stmt, void* data)
+{
+    qint64* mtime = reinterpret_cast<qint64*>(data);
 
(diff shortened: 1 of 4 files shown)


src/avatar-service.cpp
```diff
@@ -229,12 +302,14 @@ QImage AvatarService::getAvatar(const QString& email)
 {
     QImage img = loadAvatarFromLocal(email);
 
-    if (img.isNull()) {
+    // update all avatars if feature autoupdate enabled or img is null
+    if (autoupdate_db_ || img.isNull()) {
```
lins05: Each time `AvatarService::getAvatar` is called (it's called frequently when painting activities tab), a new GetAvatarRequest would be sent (if there isn't one in the queue), this is not good. We should only send a `GetAvatarRequest` either:

- We have no avatar for the current account yet, or
- We have not checked avatars for a period of time, for example 5 minutes.

lins05:
Sometimes it fails to update the avatar even if I have updated it from seahub.

Chilledheart:
@lins05 the seahub issue is posted internally and should be fixed now(?).

src/avatar-service.cpp
Chilledheart: Thanks for the advice, I will do it sometime later.

Chilledheart: cc@lins05 Set a 5-min expire time to fix this bug.

Chilledheart:
Sec, there is still a bug for the expire time checking.

PR #502, opened 2015-03-30
mac: fix copy text to pasteboard issue

Chilledheart:
for mac, qt copys many minedatas beside public.utf8-plain-text
e.g. public.vcard, which we don't want to use


Chilledheart:
before it we have:
screen shot 2015-03-30 at 3 25 26 pm
after it we have:
screen shot 2015-03-30 at 3 34 52 pm

PR #510, opened 2015-04-02
detect existing folder when cloing repos

Chilledheart:


src/ui/download-repo-dialog.cpp
```diff
@@ -188,12 +119,16 @@ bool DownloadRepoDialog::validateInputs()
...
-    if (!dir.exists()) {
-        QMessageBox::warning(this, getBrand(),
-                             tr("The folder does not exist"),
-                             QMessageBox::Ok);
-        return false;
+    sync_with_existing_ = false;
+    // exist and is non-empty?
+    if (isNonEmptyDirectory(mDirectory->text())) {
+        sync_with_existing_ = true;
+        int ret = QMessageBox::question(
```
lins05: Better use `SeafileApplet::yesOrNoBox`?

src/ui/download-repo-dialog.cpp
```diff
@@ -18,33 +19,19 @@
...
-} // namespace
-
 DownloadRepoDialog::DownloadRepoDialog(const Account& account,
                                        const ServerRepo& repo,
                                        QWidget *parent)
     : QDialog(parent),
       repo_(repo),
-      account_(account),
-      mode_(CREATE_NEW_FOLDER)
+      account_(account)
```
lins05: `sync_with_existing_` is not initialized in the ctor.

src/ui/download-repo-dialog.cpp
```diff
@@ -18,33 +19,19 @@
...
-
-    object = json_object();
-    json_object_set_new(object, "is_readonly", json_integer(repo.readonly));
-
-    info = json_dumps(object, 0);
-    QString ret = QString::fromUtf8(info);
-    json_decref (object);
-    free (info);
-    return ret;
+static bool inline isNonEmptyDirectory(const QString &dirpath) {
```
lins05: Following the project code style we should use anonymous namespace instead of static functions.

src/ui/download-repo-dialog.cpp
Chilledheart: it is done in `validateInputs`

src/ui/download-repo-dialog.cpp
Chilledheart: The functions/variables in anonymous namespace are identical to the static ones while I saw no encourage of this in [google code style](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Namespaces). In fact, I prefer to use anonymous namespace for class definition while the functions and variables are declared static, which makes code more clear that local things are local.

Chilledheart: I will change it later.

src/ui/download-repo-dialog.cpp
Chilledheart: Any difference except the requirement with the global class `SeafileApplet`?

lins05: It can reduce the code here from 5 lines to 2 lines, and the name `yerOrNoBox` shows the intent of the code more clearly.

Chilledheart: Sorry but this involves module dependency which makes code harder to be modular.

src/ui/download-repo-dialog.cpp
```diff
@@ -222,10 +158,28 @@ void DownloadRepoDialog::onDownloadRepoRequestSuccess(const RepoDownloadInfo& in
...
     QString password = repo_.encrypted ? mPassword->text() : QString();
-    int ret;
+    int ret = 0;
     QString error;
 
-    if (mode_ == MERGE_WITH_EXISTING_FOLDER) {
+    // create if not a sync with existing
+    if (!sync_with_existing_) {
+        QDir dir(worktree);
+        ret = dir.mkpath(".") ? 0 : -1;
```
lins05: `mkpath` is not available in qt4.

Chilledheart: Yes, it is. see more: http://doc.qt.io/qt-4.8/qdir.html#mkpath

lins05: oops, you're right.

Chilledheart:
updated:
screen shot 2015-04-28 at 12 06 06 pm

PR #516, opened 2015-04-15
don't show native title bar on linux

Chilledheart:
tested under OSX 10.10, Windows 7, Ubuntu 12.04, and Debian Jessie


Chilledheart:
Here is a screenshot where the bug is something like this:
screenshot from 2015-04-15 21 16 41


After this patch, everything goes okay:

Chilledheart:
After this patch, everything goes okay:
screenshot from 2015-04-15 21 22 56

PR #522, opened 2015-04-22
auto logout current account when server returns a 401 status

Chilledheart:


src/repo-service.cpp
```diff
@@ -260,6 +260,17 @@ void RepoService::onRefreshFailed(const ApiError& error)
 {
     in_refresh_ = false;
 
+    // for the new version of seafile server
+    // we may have a 401 response whenever invalid token is used.
+    // see more: https://github.com/haiwen/seahub/commit/94dcfe338a52304f5895914ac59540b6176c679e
+    // but we only handle this error here to avoid complicate code since it is
+    // general enough.
+    // TODO move this handling to all possible cases
+    if (error.type() == ApiError::HTTP_ERROR && error.httpErrorCode() == 401) {
```
lins05: No need to compare the type of the error. `httpErrorCode` only returns non-zero if the error type is `HTTP_ERROR`.

Chilledheart: Yeah, but httpErrorCode is uninitialized if it is not constructed from `ApiError::fromHttpError`. So I thought it is not safe to use directly there.

PR #526, opened 2015-04-23
don't open browser when tray icon context menu is trigger

Chilledheart:


src/ui/tray-icon.cpp
```diff
@@ -444,30 +441,6 @@ void SeafileTrayIcon::showSettingsWindow()
...
-void SeafileTrayIcon::onClick()
-{
-    if (state_ == STATE_HAVE_UNREAD_MESSAGE) {
-        viewUnreadNotifications();
-    } else {
-        showMainWindow();
-    }
-}
-
 void SeafileTrayIcon::disableAutoSync()
```
lins05: Removing all these code would cause seafile-applet does nothing when the user left clicks the tray icon (on windows).

Chilledheart: Sorry, I failed to test the patch on Windows for I though it was a quite clean fix. Anyway I will be back soon.

Chilledheart:
@lins05 re-added the deleted part. it should work properly now (tested under Windows 7)

PR #527, opened 2015-04-23
share application instance

### LINKED ISSUE #525: Seafile Client Windows
I have several clients complain that when they click the icon for seafile on their desktop that, when the seafile client is running (but minimized to the tray), the seafile client doesn't appear and simply states that an existing instance is running. It would be more intuitive for the client to appear when the desktop icon is client, similar to how Dropbox and OneDrive behave.

Chilledheart:
Closes #525.

### DIFF
--- a/src/shared-application.cpp
+++ b/src/shared-application.cpp
@@ -57,7 +57,7 @@ void messageCallback(CcnetMessage *message, void *data)
 #ifndef Q_OS_WIN32
 void *askActivateSynchronically(void * /*arg*/) {
 #else
-DWORD WINAPI askActivateSynchronically(LPVOID /*arg*/) {
+unsigned __stdcall askActivateSynchronically(void * /*arg*/) {
 #endif
     _CcnetClient* async_client = ccnet_client_new();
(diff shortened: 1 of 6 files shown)


src/shared-application.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (ccnet_processor_start((CcnetProcessor *)mqclient_proc,
+                              G_N_ELEMENTS(topics), (char **)topics) < 0) {
+        event_base_free(ev_base);
+        g_object_unref(sync_client);
+        g_object_unref(async_client);
+        return 0;
+    }
+
+    event_base_dispatch(ev_base);
+
```
lins05: Directly handle libevent doesn't seem so good to me. Maybe we could provide a rpc service in applet, and let the newly started applet instance invoke the rpc, instead of using ccnet messages? (messages are more suitable for broadcasting, or pub/sub).

The

Chilledheart: @lins05 I must admire it is a bit complicate here. What about discussing it in details next week?

PR #534, opened 2015-04-28
filebrowser: add a share current folder button

Chilledheart:
- add a workaround for servers older than <del>4.2.1</del> 4.2.0


Chilledheart:
screen shot 2015-04-28 at 3 06 01 pm

PR #541, opened 2015-04-30
utilize getSystemVersions for windows and mac os x

Chilledheart:
- next step after this is to add helpers for kinds of linux desktops


Chilledheart:
@lins05 can I merge this? it refactors the existing code and adds some convenient functions to detect system version. And then, we can add support to detect linux distribution as well!

lins05:
go ahead :)

Chilledheart:
thx!

PR #543, opened 2015-05-04
mac: use native NSUserNotification (supported by qt) if possible

Chilledheart:
it doesn't cost too many lines but uses mac native notification which is integrated with os better.

screen shot 2015-05-04 at 4 45 37 pm

(for OS X 10.10 or later)
screen shot 2015-05-04 at 4 45 58 pm


Chilledheart:
Anyway, there are some several blocking factors:
- [x] it requires a 10.8 targeted qt build (official binary has done it)
- [ ] Qt's implementation is not pragrammable  (such as control delivering time, sound and respond)
- [ ] Qt's implementation only present only one of notifications. (in fact, it should use scheduleNotification, not deliverNotification)

PR #544, opened 2015-05-04
windows: open current repo's worktree when clicking ballon message

### LINKED ISSUE #542: suggestion - clicking on tray popup info after new sync
hi guys,

If a new file have been synced bext to the tray icon comes a message..."file/folde whatever added/changed"

is it possible to implement a feature: if I click on that message that the order where the added/changed file/folder would directly open?!

I only know this feature from dropbox (sry for that ;-))

greets
fabian

Chilledheart:
Closes #542.

### DIFF
--- a/src/ui/tray-icon.cpp
+++ b/src/ui/tray-icon.cpp
@@ -16,6 +16,7 @@ extern "C" {
 #include <QDebug>
 #include <QMenuBar>
 
+#include "rpc/local-repo.h"
 #include "utils/utils.h"
 #include "utils/utils-mac.h"
 #include "seafile-applet.h"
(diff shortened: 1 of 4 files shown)


src/ui/tray-icon.cpp
```diff
@@ -101,6 +102,11 @@ SeafileTrayIcon::SeafileTrayIcon(QObject *parent)
     connect(SeahubNotificationsMonitor::instance(), SIGNAL(notificationsChanged()),
             this, SLOT(onSeahubNotificationsChanged()));
 
+#ifdef Q_OS_WIN32
+    connect(this, SIGNAL(messageClicked()),
```
lins05: What about mac?

Chilledheart: mac is different, see more http://doc.qt.io/qt-4.8/qsystemtrayicon.html#messageClicked. Qt is always lazy with mac support. 
> This signal is emitted when the message displayed using showMessage() was clicked by the user.
> Currently this signal is not sent on Mac OS X.



Anyway you will need #543 and some extra code to handle with it.

PR #546, opened 2015-05-05
fix account avatar showing up issue when starting

Chilledheart:


Chilledheart:
@lins05 please double-check this change before we can merge it!

PR #553, opened 2015-05-08
handle ssl error correctly

Chilledheart:


Chilledheart:
@lins05 can we get this merged before next release? It is not vitally important, but fixes the broken user experience with custom-CA seafile servers.

PR #556, opened 2015-05-09
network: disable weak ciphers explicitly

Chilledheart:


src/network-mgr.cpp
```diff
@@ -2,12 +2,99 @@
...
+        else if (name == "MD5")
+            return true;
+        else if (name == "DES")
+            return true;
+        else if (name == "DSS")
+            return true;
+        else if (name == "EXP")
+            return true;
+        else if (name == "NULL")
+            return true;
```
lins05: It'll be more readable to construct a `QSet<String>` and use `QSet.contains()` than this kind of lots of if..else statements.

lins05:
@Chilledheart Any reference document on which ciphers suite to use?

Chilledheart:
@lins05 Yes, I think we do have some. 

For example, in the [RFC7465](https://tools.ietf.org/html/rfc7465), RC4 cipher is found insecure and not recommended to use. For a complete list of choice of these ciphers, you can refer to https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html .

src/network-mgr.cpp
Chilledheart: @lins05 sorry, but I can't agree with you about using `QSet` and `QHashMap` everywhere. It just makes no sense to use `QSet` here but wastes a lot of CPU time if we can use plain `vector` instead. (Internally, the methods of `QSet` and `QHashMap` call hundred times of instructions than plain `vector` do, cost much more memory and don't scale well)

shoeper:
See also https://www.rfc-editor.org/rfc/rfc7525.txt

Chilledheart:
@shoeper Thanks for providing this information!

PR #558, opened 2015-05-09
[win] use preconfigure directory for initialize seafile

Chilledheart:


src/configurator.cpp
```diff
@@ -30,6 +31,16 @@ const char *kMyComputerNamespacePath =
     "Software\\Microsoft\\Windows\\CurrentVersion"
     "\\Explorer\\MyComputer\\Namespace";
 
+const char* const kPreconfigureDirectory = "PreconfigureDirectory";
+QString getPreconfigureDirectory()
+{
+    RegElement reg(HKEY_CURRENT_USER, "SOFTWARE\\Seafile", kPreconfigureDirectory, "");
```
lins05: I see the key is "SOFTWARE\\Seafile", isn't windows registry key case sensentive? (I'm not sure)

src/configurator.cpp
```diff
@@ -78,6 +89,27 @@ void Configurator::initCcnet()
...
+#if defined(Q_OS_WIN32)
+    QString preconfigure_dir = getPreconfigureDirectory();
+    if (!preconfigure_dir.isEmpty()) {
+        QDir dir(preconfigure_dir);
+        if (!dir.mkpath(".") ||
+            !dir.mkpath("Seafile/.seafile-data")) {
+            QMessageBox::warning(NULL, getBrand(),
+                                 tr("Unable to create preconfigure directory"),
+                                 QMessageBox::Ok);
+            exit(1);
```
lins05: When we failed to create the foder whose path is pre-configured in the registry, we have two options:

- show an error box and exit (as this PR does)
- ask the user to choose a directory (as we used to).

I'm not sure which one is better.

PR #567, opened 2015-05-12
fix view url on cloud view

Chilledheart:
we adopted a new format of cloud view url from server version 4.2.0

### DIFF
--- a/src/ui/repo-tree-view.cpp
+++ b/src/ui/repo-tree-view.cpp
@@ -506,7 +506,14 @@ void RepoTreeView::viewRepoOnWeb()
 {
     const Account account = seafApplet->accountManager()->currentAccount();
     if (account.isValid()) {
-        QDesktopServices::openUrl(account.getAbsoluteUrl("repo/" + selected_repo_.id));
+        // we adopt a new format of cloud view url from server version 4.2.0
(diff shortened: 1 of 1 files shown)


src/ui/repo-tree-view.cpp
```diff
@@ -506,7 +506,14 @@ void RepoTreeView::viewRepoOnWeb()
 {
     const Account account = seafApplet->accountManager()->currentAccount();
     if (account.isValid()) {
-        QDesktopServices::openUrl(account.getAbsoluteUrl("repo/" + selected_repo_.id));
+        // we adopt a new format of cloud view url from server version 4.2.0
+        if (!account.isAtLeastVersion(4, 2, 0)) {
+            QDesktopServices::openUrl(account.getAbsoluteUrl("repo/" + selected_repo_.id));
+        } else {
+            QUrl url = account.getAbsoluteUrl("/");
+            url.setFragment("common/lib/" + selected_repo_.id + "/");
```
freeplant: # is missing, like https://example.com/#common/lib/3e0f7863-89df-4c14-8f4c-644142238739

Chilledheart: `QUrl::setFragment` implies it sets up the `#` part of url.

PR #574, opened 2015-05-15
filebrowser: enhance file type

Chilledheart:


src/utils/file-utils.cpp
```diff
@@ -637,6 +637,51 @@ QString mimeTypeFromFileName(const QString& fileName)
...
+    } else if (mimetype.startsWith("audio")) {
+        return QObject::tr("Audio File");
+    } else if (mimetype.startsWith("video")) {
+        return QObject::tr("Video File");
+    } else if (mimetype.contains("msword") || mimetype.contains("ms-word")) {
+        return QObject::tr("Word Document");
+    } else if (mimetype.contains("mspowerpoint") || mimetype.contains("ms-powerpoint")) {
+        return QObject::tr("PowerPoint Document");
+    } else if (mimetype.contains("msexcel") || mimetype.contains("ms-excel")) {
+        return QObject::tr("Excel Document");
```
lins05: Shouldn't we make word/excel/ppt files close to each other when sorting by document type?

Chilledheart: @lins05 yeah, that's reasonable.

PR #581, opened 2015-05-19
improve ssl error handling output

Chilledheart:


src/utils/utils.cpp
```diff
@@ -719,22 +719,21 @@ QString dumpCertificate(const QSslCertificate &cert)
...
     s += "CommonName(CN):             " + CERTIFICATE_STR(cert.subjectInfo(QSslCertificate::CommonName)) + "\n";
     s += "Organization(O):            " + CERTIFICATE_STR(cert.subjectInfo(QSslCertificate::Organization)) + "\n";
     s += "OrganizationalUnitName(OU): " + CERTIFICATE_STR(cert.subjectInfo(QSslCertificate::OrganizationalUnitName)) + "\n";
     s += "Serial Number:              " + dumpHexPresentation(cert.serialNumber()) + "\n";
 
-    s += "\nIssued By\n";
+    s += "\nIssued By:\n";
     s += "CommonName(CN):             " + CERTIFICATE_STR(cert.issuerInfo(QSslCertificate::CommonName)) + "\n";
     s += "Organization(O):            " + CERTIFICATE_STR(cert.issuerInfo(QSslCertificate::Organization)) + "\n";
     s += "OrganizationalUnitName(OU): " + CERTIFICATE_STR(cert.issuerInfo(QSslCertificate::OrganizationalUnitName)) + "\n";
```
lins05: Another way to do it is to define `s` as a `QStringList`, and finally use `s.join("\n")` to concatenate all the lines.

Chilledheart: Yeah, but I don't tend to refactor this part since we can use `QSslCertificate::toText` method which is introduced from Qt 5.

lins05: OK. Fair enough.

PR #587, opened 2015-05-23
filebrowser: fix broken title bar when minimized on windows

Chilledheart:


Chilledheart:
How to reproduce the issue:
- open seafile-client application on windows
- open filebrowser on any repository
- click minimize button

PR #586, opened 2015-05-22
Feature file search

Chilledheart:
screen shot 2015-05-26 at 3 14 51 pm

PR #591, opened 2015-05-26
repo-service: add api to open folder in filebrowser

Chilledheart:


Chilledheart:
Test: tested with #586

PR #601, opened 2015-06-01
ui: add an option to syncing with an existing folder with a different name

Chilledheart:
-  download repo dialog: add an option to sync to an alternative path


Chilledheart:
screen shot 2015-06-01 at 2 32 56 pm

PR #602, opened 2015-06-01
ui: enhance share link dialog

Chilledheart:
screen shot 2015-06-01 at 3 02 22 pm
screen shot 2015-06-01 at 3 03 28 pm

### DIFF
--- a/src/filebrowser/sharedlink-dialog.cpp
+++ b/src/filebrowser/sharedlink-dialog.cpp
@@ -20,13 +20,18 @@ SharedLinkDialog::SharedLinkDialog(const QString &text, QWidget *parent)
     layout->setSpacing(5);
     layout->setContentsMargins(9, 9, 9, 9);
 
-    QLineEdit *editor = new QLineEdit;
-    editor->setText(text_);
-    editor->selectAll();
-    editor->setReadOnly(true);
-    layout->addWidget(editor);
+    editor_ = new QLineEdit;
+    editor_->setText(text_);
+    editor_->selectAll();
+    editor_->setReadOnly(true);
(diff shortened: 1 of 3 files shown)


shoeper:
is it really "Download Only"?

You mean "Direct Download", don't you?

Chilledheart:
@shoeper Yeah, you're correct.

src/filebrowser/sharedlink-dialog.cpp
```diff
@@ -16,18 +17,25 @@ SharedLinkDialog::SharedLinkDialog(const QString &text, QWidget *parent)
...
+    editor_->selectAll();
+    editor_->setReadOnly(true);
+    layout->addWidget(editor_);
 
     QHBoxLayout *hlayout = new QHBoxLayout;
 
+    QCheckBox *is_download_checked = new QCheckBox(tr("Direct Download"));
+    connect(is_download_checked, SIGNAL(stateChanged(int)),
+            this, SLOT(onDownloadStateChanged(int)));
+    is_download_checked->setFont(changeFontSize(is_download_checked->font(), 10));
```
lins05: Why not set the font size in style sheet?

Chilledheart: It costs at least two lines of code to set font size in style sheet (`QObject::setObjectName` and `font-size:xxx`), while it costs only one loc (line of code) to set it in `C++`.

Chilledheart: @lins05 I made some changes hoping the code is more clear.

lins05:
The ok button should be "OK" instead of "Ok".

PR #640, opened 2015-06-29
Implement get seafile link

Chilledheart:


Chilledheart:
screen shot 2015-06-29 at 4 07 52 pm

src/open-local-helper.cpp
```diff
@@ -94,6 +95,34 @@ void OpenLocalHelper::sendOpenLocalFileMessage(const char *url)
...
+        url.addQueryItem("email",  email);
+    url.addQueryItem("path",  path);
+#endif
+
+    return url;
+}
+
+QUrl OpenLocalHelper::generateLocalFileWebUrl(const QString& repo_id, const Account& account, const QString& path)
+{
+    return account.getAbsoluteUrl("/#common/lib/" + repo_id + (path.startsWith("/") ? path : "/" + path));
```
lins05: The pattern `#common/lib/` is only for folders. For files we should use the pattern `https://seafile.example.com/lib/<repo-id>/file/encoded-filename.txt`

src/open-local-helper.cpp
```diff
@@ -94,6 +95,34 @@ void OpenLocalHelper::sendOpenLocalFileMessage(const char *url)
...
+    if (!email.isEmpty())
+        url_query.addQueryItem("email",  email);
+    url_query.addQueryItem("path",  path);
+    url.setQuery(url_query);
+#else
+    url.addQueryItem("repo_id",  repo_id);
+    if (!email.isEmpty())
+        url.addQueryItem("email",  email);
+    url.addQueryItem("path",  path);
+#endif
```
lins05: There is a convenient method [includeQueryParams](https://github.com/haiwen/seafile-client/blob/v4.2.5/src/utils/utils.h#L89) which wraps the difference of qt4 and qt5.

src/repo-service-helper.cpp
```diff
@@ -57,7 +57,7 @@ void FileDownloadHelper::onCancel()
         disconnect(req_, 0, this, 0);
 }
 
-void FileDownloadHelper::onGetDirentsSuccess(const QList<SeafDirent> &dirents)
+void FileDownloadHelper::onGetDirentsSuccess(bool current_readonly, const QList<SeafDirent> &dirents)
```
lins05: the param `current_readonly` seems not used in this function?

src/filebrowser/seafilelink-dialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+    hlayout->addWidget(spacer3);
+
+    QPushButton *ok = new QPushButton(tr("OK"));
+    hlayout->addWidget(ok);
+    connect(ok, SIGNAL(clicked()), this, SLOT(accept()));
+    ok->setFocus();
+
+    layout->addLayout(hlayout);
+
+    setLayout(layout);
```
lins05: Why not use qt designer? It's much simpler and intuitive than writing code to do the ui layouts.

src/open-local-helper.cpp
Chilledheart: exactly

src/open-local-helper.cpp
```diff
@@ -94,6 +95,38 @@ void OpenLocalHelper::sendOpenLocalFileMessage(const char *url)
...
+    return url;
+}
+
+QUrl OpenLocalHelper::generateLocalFileWebUrl(const QString& repo_id, const Account& account, const QString& path)
+{
+    QString fixed_path = path.startsWith("/") ? path : "/" + path;
+    if (fixed_path.endsWith("/"))
+        return account.getAbsoluteUrl("/#common/lib/" + repo_id + fixed_path);
+    else
+        return account.getAbsoluteUrl("/lib/" + repo_id + "/file" + fixed_path);
```
lins05: `fixed_path` must be url encoded.

PR #647, opened 2015-07-03
fsplugin: implement internallink command

Chilledheart:


Chilledheart:
I am going to merge it since it is good to go.

PR #657, opened 2015-07-08
[win] Prefer global server configuration through HKLM

moobyfr:
HKCU can be used only for current logged-in user. 
HKLM will be easier because the settings will apply to all users globally, for new connected users or users which have already logged onto the workstation once

### DIFF
--- a/src/configurator.cpp
+++ b/src/configurator.cpp
@@ -34,7 +34,7 @@ const char *kMyComputerNamespacePath =
 const char* const kPreconfigureDirectory = "PreconfigureDirectory";
 QString getPreconfigureDirectory()
 {
-    RegElement reg(HKEY_CURRENT_USER, "SOFTWARE\\Seafile", kPreconfigureDirectory, "");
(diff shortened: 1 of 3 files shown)


moobyfr:
The documentation needs to be updated for using HKLM:
the registry entry key to use is 
 - \HK_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Seafile
for keys 
 - PreconfigureDirectory
 - PreconfigureServerAddr
 - PreconfigureServerAddrOnly

because the software is a 32bits application

shoeper:
Not sure if this is a good idea. Seafile client settings are imho user specific.

moobyfr:
Only preconfiguration values are stored in the HKLM tree. Other settings are stored in HKCU, which is private. Using HKLM for system-wide settings is very useful for large deployments. This will permit to deploy seafile-client on a large organisation, and prefill the server settings with an private server automatically.

Chilledheart:
I am going to merge this and improve the algorithm by checking preconfigure settings in HKCU first and then those in HKLM.
If you have any comment or feedback, please tell us asap.

PR #659, opened 2015-07-09
popup message box when open local file operation fails

Chilledheart:


src/repo-service.cpp
```diff
@@ -330,11 +330,13 @@ void RepoService::openLocalFile(const QString& repo_id,
     } else {
         ServerRepo repo = getRepo(repo_id);
         if (!repo.isValid()) {
-            qWarning("trying to open invalid repo %s", repo_id.toUtf8().data());
+            QString msg = tr("Unable to open file \"%1\" from nonexistent repo \"%2\"").arg(path_in_repo).arg(repo_id);
```
freeplant: Change repo to library

Chilledheart: okay

PR #672, opened 2015-07-16
filebrowser: implement lock support

Chilledheart:


lins05:
screen shot 2015-08-04 at 15 13 01

PR #711, opened 2015-09-08
fixed "drop here" text displaying when focused

lins05:
Germany:

screen shot 2015-09-08 at 17 53 20

Chinese:

screen shot 2015-09-08 at 17 52 11

English: 
screen shot 2015-09-08 at 17 53 54


shoeper:
I'd add some space after the box around select and `or`.

lins05:
@shoeper Makes sense. I'll improve it.

Issue #721, opened 2015-09-19
Tray icon should not change on gnome-shell


bagage:
Gnome-shell actually deprecated tray icons since 3.16 or 3.18, but there are still available in left bottom corner.

The proper way would be to use a persistent notification (see [Transmission related topic](https://trac.transmissionbt.com/ticket/3685)), but the current solution still work nowadays except the fact that the icon is dynamically changed upon time.
It cause the sys tray bar to toggle on/off when state changes.

It would be great to completely stop changing icon on gnome-shell and only keep the standard icon - if one wants to see synchronization status he/she can still open Seafile.

PR #744, opened 2015-11-11
tray-icon.cpp: do not rotation tray icon on Gnome to improve user experience (no more tray bar show/hide toggles)

bagage:
Fixing issue #721.

### DIFF
--- a/src/ui/tray-icon.cpp
+++ b/src/ui/tray-icon.cpp
@@ -212,6 +212,12 @@ void SeafileTrayIcon::createGlobalMenuBar()
 
 void SeafileTrayIcon::rotate(bool start)
 {
+    /* tray icon should not be refreshed on Gnome according to their guidelines */
+    const char *env = g_getenv("DESKTOP_SESSION");
+    if (env && strcmp(env, "gnome") == 0) {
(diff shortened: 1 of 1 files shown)


src/ui/tray-icon.cpp
```diff
@@ -260,6 +260,12 @@ void SeafileTrayIcon::showMessageWithRepo(const QString& repo_id, const QString
 
 void SeafileTrayIcon::rotateTrayIcon()
 {
+    /* tray icon should not be refreshed on Gnome according to their guidelines */
+    const char *env = g_getenv("DESKTOP_SESSION");
+    if (env && strcmp(env, "gnome") == 0) {
+        return;
+    }
+
```
lins05: Thanks for the PR! However I think the function `SeafileTrayIcon::rotate` is a better place to put the code. 

https://github.com/haiwen/seafile-client/blob/3829b0a3e731c9d0baa43c60d249422aabd81d80/src/ui/tray-icon.cpp#L213

bagage: Indeed, I moved piece of code accordingly.

PR #793, opened 2016-03-25
Qt56 better hdpi support

lins05:


lins05:
The travis ci builds failed because Qt 5.6.0 removes `QtWebKit` (superseded by `QtWebEngine`), we need to update `ShibbolethLoginView` to adapt to this change.

Issue #634, opened 2015-06-24
can't interact with system tray icon


valeth:
I can't open the context menu when right clicking the system tray icon and a left klick does not open the client window. As a result of that I have to terminate the client if I want to close it. Also, the settings menu is not acessible.

I'm using KDE Plasma version 5.3.1 on Arch Linux x86_64

dndesigns:
Experiencing the same issue. Manjaro (Arch derivative) and KDE Plasma 5.3.2.

Astate:
same probleme here

MyerSmith:
Really annoying!

seafile-client-qt5 5.0.4-1 is under plasma-workspace 5.5.4-1 (archlinux) not rightclickable but visible.

qt4 Version works but is also not usable because xembedsniproxy is a KDE bughell..

hcmh:
Is there any update on how to fix this? Since qt4 might become (or is already?) unsupported, it'd be very nice to have a functional client on KDE Plasma...

([here](https://aur.archlinux.org/packages/seafile-client/) someone claiming to be from the seafile team says that qt4 will become unsupported)

src386:
Up.
Really annoying bug on Kubuntu 16.04 with Plasma 5...
I need to pkill seafile then start it each time I want to interact with it. And it seems impossible to access the parameters...

nidico:
I just ran into that issue on kubuntu 16.04 and it's a major dealbreaker in our company, so it'd be great if this could be fixed soon!

Astate:
Same problem here
Le 20 mai 2016 07:42, "Nicolas Dietrich" <notifications@github.com> a
écrit :

nidico:
FYI: It's also broken on Plasma 5.6 and Qt 5.6 (from KDE neon repositories) on top of kubuntu xenial.

lins05:
I tried the latest fedora 23 (KDE5 + qt5.6), and it seems to work without a problem.

![e](https://cloud.githubusercontent.com/assets/717363/15571993/aeb25b72-2370-11e6-82b4-38c224db4331.png)

I'll check ubuntu 16.04 + KDE5 later.

lins05:
Fxied in https://github.com/haiwen/seafile-client/pull/802

nidico:
Confirmed - works nicely in 5.1.2.

PR #802, opened 2016-05-27
Regression fix for StatusNotifierItem protocol

dbruenig:
The removed lines were causing a problem for the context menu to not work with qt5.6 which uses DBus, SNI and the com.canonical.dbusmenu interface. Apparently this cannot just be related to seafile but the hide() and immediately following show() call  somehow caused the Menubar object of the StatusNotifierItem to 'disappear' effectively removing the dbus  context menu object.


freeplant:
Thanks. We will give it a try.

lins05:
Thanks for the patch!

Tested on Ubuntu 16.04 with KDE neon, and it does fix the problem described in #634.

Also tested on ubuntu 14.04 with qt5.5 + unity, and it works fine.

PR #807, opened 2016-06-13
added files for debian packaging

unDocUMeantIt:
- the directory debian was added
  - it can be used for proper packaging, building the binary package
    seafile (desktop applet)
  - this is what enabled the PPA https://launchpad.net/~m.eik/+archive/ubuntu/seafile-experimental


debian/control
```diff
@@ -0,0 +1,46 @@
+Source: seafile-gui
+Section: net
+Priority: extra
+Maintainer: m.eik michalke <meik.michalke@hhu.de>
+Build-Depends:
+    debhelper (>= 7),
+    dh-python,
+    python,
```
lins05: `Seafile-client` has no python modules, so I guess we can remove `dh-python` and `python` from the build deps?

debian/control
```diff
@@ -0,0 +1,46 @@
...
+    autotools-dev,
```
lins05: I guess `autotools-dev` is also not required since we're using cmake here.

debian/rules
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+
+%:
+	dh $@ --with python2 --builddirectory=build
```
lins05: ditto for python

debian/rules
```diff
@@ -0,0 +1,12 @@
...
+
+override_dh_auto_configure:
+	export QT_SELECT=5
+	dh_auto_configure -- -DUSE_QT5=ON -DCMAKE_BUILD_TYPE=Release -DBUILD_SHIBBOLETH_SUPPORT=ON
```
lins05: `USE_QT5` has been removed for a while (we dropped qt4 and only supports qt5 since 5.1.0).

debian/control
unDocUMeantIt: aren't the two python scripts in the scripts directory (build_helper.py and build.py) used for the build process?

lins05: They're only used in building the Mac OSX package.

unDocUMeantIt: done.

debian/control
unDocUMeantIt: done.

debian/rules
unDocUMeantIt: done.

debian/rules
unDocUMeantIt: done.

PR #821, opened 2016-07-29
Replace Arial with Open Sans

schokonsti:
Use Open Sans as standard Interface font and promote free software. Fallback to Sans-Serif fonts if Open Sans is not available.


jhasse:
What about Liberation Sans? It has the same metrics as Arial and is also free.

PR #827, opened 2016-09-23
Update tray-icon.cpp to also ignore gnome on Wayland

bagage:


bagage:
@lins05 any chance to have this merged before next release?

src/ui/tray-icon.cpp
```diff
@@ -268,7 +268,7 @@ void SeafileTrayIcon::rotate(bool start)
 {
     /* tray icon should not be refreshed on Gnome according to their guidelines */
     const char *env = g_getenv("DESKTOP_SESSION");
-    if (env && strcmp(env, "gnome") == 0) {
+    if (env && (!strcmp(env, "gnome") || !strcmp(env, "gnome-wayland"))) {
```
lins05: Please use `strcmp(a, b) == 0` instead of negating it to make it more readable.

Issue #809, opened 2016-06-16
Huge client window on gnome shell


woprandi:
Hi,

I use the seafile client on Fedora 23 (GNOME Shell) with this repo : https://copr.fedorainfracloud.org/coprs/pkerling/seafile/

Since the version 5.1.1 (this issue was not present on 5.0.7), the client is appears "huge" (as if I zoomed)
![image](https://cloud.githubusercontent.com/assets/1449422/16108348/e1a0f074-33a2-11e6-9497-b82fbb7a64c4.png)

With multi monitors, I have not this issue if I drag the window on a external monitor.

shoeper:
I do have the same issue on Fedora 23

woprandi:
After upgrading to Fedora 24. The issue is still there

FeuRenard:
I have the same issue on Fedora 24, GNOME shell. It's annoying, as there is no possibility to resize the window. So the indicators for disk space in the cloud and up-/download cannot be shown.

FeuRenard:
I set up a second system with Fedora 24 and GNOME shell, but the window had a comfortable size.

1. Notebook, 13" display, 1920x1080 => window too large
2. Desktop, 22" display, 1680x1050 => window with comfortable size

I suppose, that this issue correlates with the used screen size. Can you two support this claim, maybe with your screen sizes and resolutions?

shoeper:
14" Lenovo Thinkpad Laptop

woprandi:
@FeuRenard I confirm that.

On external monitors (~22/23", 1920x1080) -> good size
On laptop screen (14", 1920x1080) -> issue

gnacho:
Same problem here :(

Loris1123:
I'm also having this problem on the latest XFCE4. But a friend, using xfce aswell, doesn't have problem.

I'm running Manjaro Linux but switched to xfce a few days ago. I installed/compiled seafile on gnome-shell.

wederer:
I have the same issue on my laptop (1080p) running arch with i3.

jhasse:
On my computer VLC also has this issue, it seems to be something Qt related.

jhasse:
Adding

```
export QT_AUTO_SCREEN_SCALE_FACTOR=0.8
```

to `~/.profile` and rebooting fixed it for me :) Can others confirm?

brege:
jhasse: Good work around!  Can confirm it is working for me, both with Seafile and VLC.

[...]

PR #858, opened 2017-01-21
Don't use QT5.6 autoscaling on Linux

flexo3001:
The autoscaling feature since Qt 5.6 applies a 2x scaling at >144 DPI. This affects not just users with real hiDPI displays since a display with a fullhd resolution and a <=15.2" size has a greater DPI than the default threshold. Because of the fact that e.g. GNOME starts scaling at 192 DPI and everything below needs to be done manually it's much saner to disable the autoscale feature until more people own a real hiDPI display. This should 'fix' #809 for now.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -64,7 +64,7 @@ void setupHIDPIFix()
     qApp->setAttribute(Qt::AA_UseHighDpiPixmaps);
 #endif
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) && !defined(Q_OS_MAC)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) && defined(Q_OS_WIN)
     // Enable HDPI auto detection.
     // See http://blog.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6/
     qApp->setAttribute(Qt::AA_EnableHighDpiScaling);


shoeper:
This is no solution as well.

flexo3001:
I know, i see the need of a functioning scaling as well but right now it's unusable.

shoeper:
On which devices? I didn't have any issue on a windows device so far.

flexo3001:
For example all laptops with a screen <15" and fullHD resolution.

shoeper:
Running which operating system?


Am 21. Januar 2017 20:34:48 schrieb Marco Kundt <notifications@github.com>:

flexo3001:
On GNU/Linux. I saw this bug on Arch Linux and Fedora because they are the first distros which adopted the new QT version. I can't say anything about Ubuntu 16.10 but it's possible as they are serving QT 5.6.

shoeper:
There is the problem. While it does not work on Linux it does work on Windows. And while I also like using Linux there are many more Windows users for sure and thus I think this option should be kept enabled because most users have an advantage. Maybe it could be disabled for Linux/UNIX only using a compiler statement / macro.

lins05:
I'd suggest keeping the current behaviour, and use an environment variable to control this. E.g. `SEAFILE_DISABLE_QT_HDPI`.

shoeper:
The factor can also be influenced by `export QT_AUTO_SCREEN_SCALE_FACTOR=1`. It can be added to `~/.profile` to be loaded on login. I've set it to 0.5 and it works great.

flexo3001:
I don't care anymore. Fedora patched QT5 to a much saner threshold. For my sake close this or not.

Issue #785, opened 2016-03-14
Strange behaviour of progress


FlorianLoch:
I noticed some strange behaviour regarding the progress bar during upload of single file via "Cloud file browser":

It starts quite strange, than seems to work fine. Although fails in another case (the first file seemed to be too big for the server anyways...)

<img width="327" alt="bildschirmfoto 2016-03-10 um 08 28 01" src="https://cloud.githubusercontent.com/assets/5587805/13751998/0a132098-ea0d-11e5-934e-752fac89db31.png">
<img width="330" alt="bildschirmfoto 2016-03-10 um 08 47 22" src="https://cloud.githubusercontent.com/assets/5587805/13751999/0a133fd8-ea0d-11e5-9347-72b6dced6a33.png">
<img width="333" alt="bildschirmfoto 2016-03-10 um 22 59 39" src="https://cloud.githubusercontent.com/assets/5587805/13752000/0a1350c2-ea0d-11e5-85c1-1e785a09deb1.png">

This might be related to #751.

d0teater:
Noticed this too on the Windows client. Works fine, but progress bar isn't accurate:

<img width="182" alt="seafile_counter" src="https://cloud.githubusercontent.com/assets/10689351/23528227/4cf3ee42-ff56-11e6-9f2c-037f85006f48.png">
<img width="191" alt="seafile_counter2" src="https://cloud.githubusercontent.com/assets/10689351/23528228/4cf4cf2e-ff56-11e6-9ccf-c5bd172cbb10.png">

lins05:
Should be fixed in #898

PR #898, opened 2017-04-26
Fix integer conversion overflow

yuyichao:
In `FileBrowserProgressDialog::onProgressUpdate`. There might be more professional ways to compute the ratios but going though a floating point is the simplest I can think of...

Haven't tested yet since I don't want to abort my 4.4GB file upload.

Possibly fix ~~(don't autoclose it yet github)~~ https://github.com/haiwen/seafile-client/issues/785

### DIFF
--- a/src/filebrowser/progress-dialog.cpp
+++ b/src/filebrowser/progress-dialog.cpp
@@ -5,6 +5,7 @@
 #include <QPushButton>
 #include <QDesktopServices>
 #include <QDebug>
+#include <climits>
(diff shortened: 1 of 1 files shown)


yuyichao:
Local tests seem to work.

yuyichao:
Ping.

freeplant:
Thank you. We will look into it.

lins05:
@yuyichao thanks for the patch!

PR #918, opened 2017-06-07
Fix Qt Version Detection

moschlar:
Explicitly calling qmake to determine the Qt version is not really
necessary and probably bad style.
Still needs to be tested on Windows and Mac OS though.


moschlar:
Closing like #1224 due to #1300

PR #924, opened 2017-06-15
Added empty folder view in cloud file browser.

haidian6672:
当 cloud file browser 目录内容为空时，显示一行提示信息 “This folder is empty.”

-add-spacer

### DIFF
--- a/src/filebrowser/file-browser-dialog.cpp
+++ b/src/filebrowser/file-browser-dialog.cpp
@@ -37,10 +37,12 @@ namespace {
 enum {
     INDEX_LOADING_VIEW = 0,
     INDEX_TABLE_VIEW,
-    INDEX_LOADING_FAILED_VIEW
+    INDEX_LOADING_FAILED_VIEW,
+    INDEX_EMPTY_VIEW
 };
 
 const char *kLoadingFaieldLabelName = "loadingFailedText";
+const char *kEmptyFolderLabelName = "emptyFolderText";
 const int kToolBarIconSize = 20;
 const int kStatusBarIconSize = 24;
 //const int kStatusCodePasswordNeeded = 400;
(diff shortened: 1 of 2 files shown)


lins05:
What about moving the text to the vertical center of the dialog?

haidian6672:
How about this?
-v-center

src/filebrowser/file-browser-dialog.cpp
```diff
@@ -427,6 +431,22 @@ void FileBrowserDialog::createLoadingFailedView()
...
+    QVBoxLayout *layout = new QVBoxLayout;
+    empty_view_->setLayout(layout);
+
+    QLabel *label = new QLabel;
+    label->setObjectName(kEmptyFolderLabelName);
+    QString label_text = tr("This folder is empty.");
+    label->setText(label_text);
+    label->setAlignment(Qt::AlignCenter);
+
+    layout->addWidget(label);
```
lins05: Why creating a parent widget + layout + QLabel? I think a QLabel is all we need.

src/filebrowser/file-browser-dialog.cpp
```diff
@@ -427,6 +431,22 @@ void FileBrowserDialog::createLoadingFailedView()
...
 
+void FileBrowserDialog::createEmptyView()
+{
+    empty_view_ = new QWidget(this);
+
+    QVBoxLayout *layout = new QVBoxLayout;
+    empty_view_->setLayout(layout);
+
+    QLabel *label = new QLabel;
+    label->setObjectName(kEmptyFolderLabelName);
```
lins05: Why do we need to set the object name?

haidian6672:
Does QWidget or QVBoxLayout in loading_view and loading_failed_view need to keep?

https://github.com/haiwen/seafile-client/blob/master/src/filebrowser/file-browser-dialog.cpp#L409
https://github.com/haiwen/seafile-client/blob/master/src/ui/loading-view.cpp#L11

lins05:
@haidian6672 Yeah, we can improve those as well.

PR #925, opened 2017-06-16
Disabled show_sync_errors menu item in tray icon when there is no sync error.

haidian6672:
Disabled show_sync_errors menu item in tray icon when there is no sync error.

### DIFF
--- a/src/ui/tray-icon.cpp
+++ b/src/ui/tray-icon.cpp
@@ -232,8 +232,8 @@ void SeafileTrayIcon::prepareContextMenu()
     }
 
     std::vector<SyncError> errors;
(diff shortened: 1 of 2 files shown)


src/ui/tray-icon.cpp
```diff
@@ -230,6 +230,19 @@ void SeafileTrayIcon::prepareContextMenu()
         enable_auto_sync_action_->setVisible(true);
         disable_auto_sync_action_->setVisible(false);
     }
+
+    std::vector<SyncError> errors;
+    int ret = seafApplet->rpcClient()->getSyncErrors(&errors, 0, 1);
```
lins05: The return type of `rpcClient()->getSyncErrors(..)` bool, not int.

PR #931, opened 2017-06-20
Add leave share menu item

haidian6672:


haidian6672:
default

haidian6672:
default

src/ui/repo-tree-view.cpp
```diff
@@ -428,6 +432,12 @@ void RepoTreeView::createActions()
 
     connect(open_in_filebrowser_action_, SIGNAL(triggered()), this, SLOT(openInFileBrowser()));
 
+    unshare_action_ = new QAction(tr("&Leave share"), this);
+    unshare_action_->setIcon(QIcon(":/images/leave-share.png"));
+    unshare_action_->setStatusTip(tr("unshare this library"));
```
lins05: "unshare this library"  is improper here. Just use "leave share".

src/ui/repo-tree-view.cpp
```diff
@@ -571,6 +581,32 @@ void RepoTreeView::shareRepoToGroup()
...
 
+void RepoTreeView::unshareRepo()
+{
+    const Account account = seafApplet->accountManager()->currentAccount();
+    const QString repo_id = selected_repo_.id;
+    const QString from_user = selected_repo_.owner;
+    UnshareRepoRequest* request =
+        new UnshareRepoRequest(account, repo_id, from_user);
+
+    connect(request, SIGNAL(success()),
```
lins05: Should we show an error message to the user when the operation fails?

shoeper:
Is there a confirmation dialog? I suggest to add one if not because clicking on a menu item can happend by accident easily. ` No` should be the default of the dialog.

lins05:
@shoeper That makes sense. 

@haidian6672  Let's ask the user to confirm before sending the request.

src/ui/repo-tree-view.cpp
```diff
@@ -583,6 +583,12 @@ void RepoTreeView::shareRepoToGroup()
 
 void RepoTreeView::unshareRepo()
 {
+    if (!seafApplet->yesOrNoBox(
+            tr("Are you sure to leave share \"%1\"?").arg(
```
shoeper: Are you sure you want to leave the share

src/ui/repo-tree-view.cpp
```diff
@@ -607,6 +615,11 @@ void RepoTreeView::onUnshareSuccess()
     }
 }
 
+void RepoTreeView::onUnshareFailed(const ApiError&error)
+{
+    seafApplet->warningBox(tr("Leave share failed"), this);
```
shoeper: Leaving share failed

PR #934, opened 2017-06-21
Set default value when get config failed in class SeafileRpcClient.

haidian6672:


haidian6672:
log msg:
[06/09/17 18:24:28]Unable to get config value download_limit: Config not exists
[06/09/17 18:24:28]Unable to get config value upload_limit: Config not exists

set default value when get config value failed.

src/rpc/rpc-client.cpp
```diff
@@ -27,6 +27,11 @@ extern "C" {
 
 namespace {
 
+enum {
+    DOWNLOAD = 0,
+    UPLOAD
```
shoeper: `= 0` ?

haidian6672: Every name in an enumeration is assigned an integral value that corresponds to its place in the order of the values in the enumeration. By default, the first value is assigned 0, the next one is assigned 1, and so on, but you can explicitly set the value of an enumerator.

https://msdn.microsoft.com/en-us/library/2dzy4k6e.aspx

haidian6672: https://github.com/haiwen/seafile-client/pull/934/files#diff-fe09a831ef666aaf1edd1508f235977eL325

false == 0 == DOWNLOAD
 true == 1 == UPLOAD

shoeper: Thanks.

PR #935, opened 2017-06-22
fix a translation error

baixiangcpp:
fix a translation error


haidian6672:
Thank you. I've updated this translation in seafile-client and seadrive-gui via Transifex.

lins05:
Translation updated. Thanks @baixiangcpp and @haidian6672

PR #940, opened 2017-06-28
fix build with openssl 1.1

### LINKED ISSUE #899: Compile issue with openssl-1.1
[ccnet-server](https://github.com/haiwen/ccnet-server), [ccnet](https://github.com/haiwen/ccnet) and [seafile](https://github.com/haiwen/seafile) compile fine with openssl-1.1 using the master branch. Would it possible that the next release of the other seafile packages also compile fine with openssl-1.1?

FauxFaux:
Fix #899 and https://bugs.debian.org/859716

Builds fine in a sid chroot.

`SSLeay_add_all_algorithms` is not required as the operation is done automatically.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -4,7 +4,7 @@ Priority: extra
 Maintainer: m.eik michalke <meik.michalke@hhu.de>
 Build-Depends:
     debhelper (>= 7),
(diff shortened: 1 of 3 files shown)


globin:
This does not build with OpenSSL older than 1.1.0

PR #950, opened 2017-07-17
Adjusted ui.

haidian6672:
Windows 10:
w
Ubuntu:


shoeper:
I like what I see.

It would also be helpful if the account changer would be more visible. Most users don't recognize that it isn't possible to switch the account.

PR #963, opened 2017-07-28
Adjusted load more button.

haidian6672:
load-more


haidian6672:
Reference link:
[http://www.qtcentre.org/threads/10110-QTableView-and-problems-with-align?p=53744#post53744](url)

lins05:
LGTM. @freeplant Do you think we should highlight the button when mouse hovering it?

freeplant:
Highlight is not needed for now.

PR #979, opened 2017-08-24
Check input path existence in DownloadRepoDialog.

haidian6672:
image

### DIFF
--- a/src/ui/download-repo-dialog.cpp
+++ b/src/ui/download-repo-dialog.cpp
@@ -275,43 +275,47 @@ bool DownloadRepoDialog::validateInputs()
         }
     }
 
-    return validateInputDirectory();
+    if (!validateInputDirectory()) {
+        return false;
+    }
 
-    sync_with_existing_ = false;
-    alternative_path_ = "";
+    if (!manual_merge_mode_) {
+        sync_with_existing_ = false;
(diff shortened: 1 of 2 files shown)


src/ui/download-repo-dialog.cpp
```diff
@@ -275,9 +275,7 @@ bool DownloadRepoDialog::validateInputs()
         }
     }
 
-    if (manual_merge_mode_) {
-        return validateInputsManualMergeMode();
-    }
+    return validateInputDirectory();
```
lins05: I don't think it's correct to return here and skip all the logic below.

src/ui/download-repo-dialog.cpp
```diff
@@ -275,45 +275,47 @@ bool DownloadRepoDialog::validateInputs()
         }
     }
 
-    if (manual_merge_mode_) {
-        return validateInputsManualMergeMode();
+    if (!validateInputDirectory()) {
+        return false;
     }
 
-    sync_with_existing_ = false;
```
lins05: You should not move the large blocks of code below - it's very difficult to identify the real changes. Do something like

```cpp
if (manual_merge_mode_) {
    return true;
}

// The original code would be unchanged
// ...
```

PR #980, opened 2017-08-26
Add more check onGetDirentsPaste()

haidian6672:


src/filebrowser/file-browser-dialog.cpp
```diff
@@ -1164,11 +1173,27 @@ void FileBrowserDialog::setFilesToBePasted(bool is_copy, const QStringList &file
...
-        current_path_ == dir_path_to_be_pasted_from_) {
-        seafApplet->warningBox(tr("Cannot paste files from the same folder"), this);
-        return;
+    if (repo_id_to_be_pasted_from_ == repo_.id) {
+        if (current_path_ == dir_path_to_be_pasted_from_) {
+            seafApplet->warningBox(tr("Cannot paste files from the same folder"), this);
+            return;
+        }
+
+        if (file_names_to_be_pasted_.isEmpty()) {
```
lins05: No need to check this. If file_names_to_be_pasted_, the paste menu item would not be visible. 

See https://github.com/haiwen/seafile-client/blob/v6.1.0/src/filebrowser/file-table.cpp#L436

src/filebrowser/file-browser-dialog.cpp
```diff
@@ -68,6 +68,15 @@ bool inline findConflict(const QString &name, const QList<SeafDirent> &dirents)
     return found_conflict;
 }
 
+QString appendTrailingSlash(const QString& input) {
+    if (!input.endsWith('/')) {
+        QString ret = input;
+        return ret.append('/');
```
lins05: Why do we need this `ret` variable?

The function can be a one-liner like  `input.endsWith("/") ? input : input + "/"`

src/filebrowser/file-browser-dialog.cpp
```diff
@@ -1164,11 +1173,27 @@ void FileBrowserDialog::setFilesToBePasted(bool is_copy, const QStringList &file
...
+
+        if (file_names_to_be_pasted_.isEmpty()) {
+            seafApplet->warningBox(tr("The pasted files do not exist"), this);
+            return;
+        }
+
+        for (const QString& name : file_names_to_be_pasted_) {
+            const QString file_path_to_be_pasted =
+                appendTrailingSlash(::pathJoin(dir_path_to_be_pasted_from_, name));
+            if (appendTrailingSlash(current_path_).startsWith(file_path_to_be_pasted)) {
```
lins05: Please add a comment/example here: 

```
// Paste /a/ into /a/b/ is not allowed
```

src/filebrowser/file-browser-dialog.cpp
haidian6672: But "Ctrl+v" will emit direntPaste() signal.

PR #988, opened 2017-10-08
Accept close event instead of ignoring it

eplightning:
Ignoring main window close event causes weird behavior as described in the forum post: https://forum.seafile.com/t/plasma-logout-canceled-by-seafile-applet/3283

VLC used to have similar issue: https://trac.videolan.org/vlc/ticket/4606

This PR replaces ignore() call with accept() which fixes that issue. I haven't noticed any problems with exit-to-tray feature. Tested on Arch Linux.

### DIFF
--- a/src/ui/main-window.cpp
+++ b/src/ui/main-window.cpp
@@ -112,7 +112,7 @@ void MainWindow::hide()
 
 void MainWindow::closeEvent(QCloseEvent *event)
 {
-    event->ignore();
+    event->accept();
     hide();
 }
 


bionade24:
What's now with this patch and[ its related forum thread](https://forum.seafile.com/t/plasma-logout-canceled-by-seafile-applet/3283/20)?

moschlar:
FYI: I've added the patch to the package in the Debian repositories.

bionade24:
Perfect I'll test it. If working, I'll make a pull request with the new version.

freeplant:
@all Thank you. We will include it in the next release.

Issue #965, opened 2017-08-01
Client is attempting to connect to 127.0.0.1


jeffmikels:
My client window shows a red chain icon in the lower left... clicking on it shows three accounts. One of them is for 127.0.0.1. I want to delete this account; however, when I click on the avatar icon in the top left, the invalid account does not show up.

![screen shot 2017-08-02 at 4 04 24 pm](https://user-images.githubusercontent.com/3506020/28892789-1bdcf9e2-779d-11e7-8171-27f0d0424fb1.png)

jeffmikels:
The 127.0.0.1 account does not exist in the accounts.db file.

jeffmikels:
Here's the relevant log entry:

```bash
logs/applet.log:[08/02/17` 15:43:05][api] network error for http://127.0.0.1:8000/media/avatars/8/8/2ac5005b0a5a8099895a2534a06f1d/resized/42/e094ee63ca52e4cd90173474238ea2be.png: Connection refused
```

freeplant:
After analysing the source code, we think the problem caused by that the server doesn't have correctly set the SERVICE_URL and FILE_SERVER_ROOT. Can you check it?

The default SERVICE_URL is 127.0.0.1:8000

jeffmikels:
Both were properly set in the web settings, but I have now added SERVICE_URL to the seahub_settings.py file.

Nonetheless, the client still attempts to connect to 127.0.0.1.

wxr9:
can you tell me how to compile the seafile-client? i have installed qt5 mingw

zix99:
This was happening to me, and I suspect because I was using `proxy_pass` in my nginx config rather than fastcgi.  Either way, I added `X-Real-IP` to my configuration, and that seemed to solve it. I also made sure my `SERVICE_URL` and `FILE_SERVER_ROOT` were set correctly. I haven't done the digging to see if it's some combination of them.

I removed much of the non-applicable parts:
```
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

location / {
        proxy_pass http://127.0.0.1:8000;
        proxy_redirect default;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header Host $host;
}
```

F0cks:
Same problem here with apache server and Ubuntu client:
```
[...]

PR #1015, opened 2017-11-12
Ignore localhost server disconnected status.

haidian6672:
https://github.com/haiwen/seafile-client/issues/965


shoeper:
And what in case I do connect to a server via localhost? Why is that invalid?

Bjarno:
@shoeper A possible workaround is check whether `127.0.0.1` exists in the `accounts.db` file. However, just removing `127.0.0.1` isn't enough, for me it displays `localhost`.

It's a workaround... So I'm not sure that this should be merged.

PR #1022, opened 2017-11-19
Initialized a pointer to zero, which was missing.

Gezzo42:


lins05:
It's already fixed in https://github.com/haiwen/seafile-client/commit/b55fa3434, thanks for the patch anyway!

PR #1028, opened 2017-11-25
Skip or Retry when upload faild

sunyuzhao:


shoeper:
Does it also support the continue option the browser upload provides? I think that would be very helpful, especially when uploading very large files and the only reason for the failure is that the token expired in the meantime but the whole file already was uploaded.

PR #1084, opened 2018-04-24
Adjust the label display in search-bar

sunyuzhao:

### DIFF
--- a/qt.css
+++ b/qt.css
@@ -130,8 +130,9 @@ QLineEdit#mSearchBar {
 }
 
 QLineEdit#mSearchBar QLabel {
-    min-width: 90px;
-    max-width: 90px;
+    text-align: center;
+    margin-left: 16px;
+    margin-right: 16px;
 }
 
 QLineEdit#mSearchBar:focus {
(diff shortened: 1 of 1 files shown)


shoeper:
Why do you set min width = max width? Depending on the language the length of the string can largely differ.

I suspect that the text is left aligned thus is would be the best to have the box as small as possible but as large as needed with the maximum being just a few pixels smaller than the whole box.

qt.css
```diff
@@ -130,8 +130,8 @@ QLineEdit#mSearchBar {
 }
 
 QLineEdit#mSearchBar QLabel {
-    min-width: 80px;
-    max-width: 80px;
+    min-width: 90px;
+    max-width: 90px;
```
shoeper: Why do you set min width = max width? Depending on the language the length of the string can largely differ.

I suspect that the text is left aligned thus is would be the best to have the box as small as possible but as large as needed with the maximum being just a few pixels smaller than the whole box.

omkarnathsingh: min should be 80 and max 100 (or even large)?

shoeper: Unfortunately I cannot test it, but would expect better results with something like

```
    text-align: center;
    margin-left: 16px;
    margin-right: 16px;
```

sunyuzhao: Thanks for advice, I will test it and look forward to better results

PR #1092, opened 2018-05-24
Fixed typo

moschlar:


src/rpc/local-repo.cpp
```diff
@@ -159,14 +159,14 @@ void LocalRepo::translateSyncError(const QString &error)
     } else if (error == "Failed to start upload.") {
         sync_error_str = QObject::tr("Failed to start upload");
 
-    } else if (error == "Error occured in upload.") {
-        sync_error_str = QObject::tr("Error occured in upload");
+    } else if (error == "Error occurred in upload.") {
```
lins05: This is hard-coded message sent from the daemon, See https://github.com/haiwen/seafile/blob/v6.1.8/daemon/sync-mgr.c#L777 so it should not be modified.

moschlar:
I've removed the changes to the fiels in `i18n/*`

lins05:
Thanks @moschlar !

Issue #1097, opened 2018-05-29
unable to compile with Qt 5.11 (qt5_use_modules in CMakeLists.txt needs to be replaced)


macearl:
In Qt 5.11 `qt5_use_modules` was removed (it was deprecated a long time ago) and should be replaced with `target_link_libraries`

However just changing the Lines containing `qt5_use_modules` in CMakeLists.txt does not seem to be enough as then there are other compile errors

ignitedPotato:
The use of `target_link_libraries` also is discouraged by Qt docs: https://doc.qt.io/qt-5/cmake-manual.html#imported-targets

The CMakeLists.txt really could use some more work..

scachemaille:
it's not how Interpret the docs.
they talk about to use imported target names instead of variable like as Qt5<Module>_LIBRARIES inside target_link_libraries. not the it's discouraged.

konstantingoretzki:
Would be great to see the fix #1099 merged, if it works.

fuelflo:
+1

fuelflo:
#1099 has been merged and works, so this issue can be closed i guess.

PR #1099, opened 2018-05-30
Fix for Qt 5.11

ignitedPotato:
As @macearl already stated in #1097 the client won't compile with Qt 5.11. My changes seem to fix the issues.


fuelflo:
@lins05 could you please merge this fix, without it we're unable to compile the seafile-client (gui) with the current Qt version.

lins05:
Thanks @ignitedPotato !

PR #1105, opened 2018-06-29
Clients should not send api2/events requests if open

sunyuzhao:


lins05:
Fixed in https://github.com/haiwen/seafile-client/commit/06df33b9

PR #1127, opened 2018-08-24
2FA: use OK as default button

### LINKED ISSUE #1126: Small issue with 2FA popup on Windows Seafile client
I've just been trying to install Seafile for the first time, and ran into a really trival, silly but non-obvious problem with the Windows client (v6.2.4). I had set up 2FA successfully for my account using the Seahub web interface, but when I tried to login with the Windows app, it would (as expected) prompt me for a 2FA token, but then the 2FA window would disappear and leave me sitting at the main login window again. Nothing seemed to be happening, so I couldn't log in. 

It took me much longer than I'd like to admit to realise that what was happening was simply that the popup dialog window has the "Cancel" button set as the default - I had been typing in the 2FA code and naturally hitting return afterwards, and since "Cancel" was the default, that was the action triggered by hitting return. So it was just dumping me back to the previous screen rather than trying to log me in.

I think this would be a really simple change to make and would stop anyone else wasting their time trying to debug something that was actually working fine!

shoeper:
I've got no dev env but according to http://doc.qt.io/archives/qt-4.8/qpushbutton.html#default-prop this should fix #1126

### DIFF
--- a/ui/two-factor-dialog.ui
+++ b/ui/two-factor-dialog.ui
@@ -77,6 +77,9 @@
        <property name="text">
         <string>OK</string>
        </property>
+       <property name="default">
+        <bool>true</bool>
+       </property>
       </widget>
      </item>
     </layout>


shoeper:
@freeplant

freeplant:
Thanks @shoeper

PR #1158, opened 2018-12-10
minor fix - added missing word "Please try [and] resync the library"

yatesco:


yatesco:
Hi @freeplant - as a native English speaker I have to disagree :-) - "Please try and resync the library" or "Please try to resync the library" are the correct form.

freeplant:
We will use "Please try to resync the library". Actually the translation files are generated automatically via a third party translation platform. They should not be modified manually. We will modify it by our own.

PR #1191, opened 2019-05-06
[fix] fix clean sync token failed when exit program

sun20121818:


src/seafile-applet.cpp
```diff
@@ -219,9 +219,9 @@ SeafileApplet::~SeafileApplet()
     delete settings_dialog_;
     delete message_poller_;
     delete rpc_client_;
-    delete daemon_mgr_;
-
     delete account_mgr_;
+
+    delete daemon_mgr_;
```
lins05: 加一行注释吧，说明为什么要先销毁 `account_mgr_` 再销毁 `daemon_mgr_`

PR #1213, opened 2019-08-16
Fix typo

moschlar:


src/ui/settings-dialog.cpp
```diff
@@ -106,7 +106,7 @@ void SettingsDialog::updateSettings()
         I18NHelper::getInstance()->setPreferredLanguage(mLanguageComboBox->currentIndex());
     }
 
-    if (language_changed && seafApplet->yesOrNoBox(tr("You have changed languange. Restart to apply it?"), this, true))
+    if (language_changed && seafApplet->yesOrNoBox(tr("You have changed language. Restart to apply it?"), this, true))
```
personalcomputer, line 109: @moschlar This is a translation string referenced by 10s of files in i18n, each one would need to be modified as well. (or you can just modify the english translation of this string)

moschlar: Since I did this via the GitHub Web GUI, I'm not motivated to do that...

PR #1279, opened 2020-03-31
Support delete multi sync sync error

sun20121818:


src/ui/sync-errors-dialog.cpp
```diff
@@ -175,6 +175,12 @@ void SyncErrorsTableView::contextMenuEvent(QContextMenuEvent *event)
 
     SyncError error = model->errorAt(row);
 
+    QModelIndexList selected = selectionModel()->selectedRows();
+    sync_errors_.clear();
```
killing: 这个改为 selected_sync_errors 吧

sun20121818: ok

PR #1288, opened 2020-04-25
fix #1287 View on cloud ends up on 404

### LINKED ISSUE #1287: "View on cloud" ends up on 404 page when server address includes port number
Hi!

Steps to reproduce:
* Seahub instance is running on non-standard HTTP port (specifically `8000`, which is the default).
* In `seafile-applet`, user right-clicks any library and chooses "View on cloud" from context menu.
* Browser is being opened on `http://_address_:8000/client-login/?next=%3A8000%2Flibrary%2F_uuid_%2F_name_%2F&token=_token_` and then redirected to `http://_address_:8000/client-login/:8000/library/_uuid_/_name_/`.
* The later says "Sorry, but the requested page could not be found".

I believe that the port part should not be in "next" query parameter.

mike386:


mike386:
Looks like it is already "merged" without accepting the PR.

PR #1297, opened 2020-05-29
Fix build with Qt5 v5.15.0

Joffr3y:


yopito:
fine for me: this patch made me able to build seafile-client 7.0.8 with qt5.15.0 (VoidLinux distribution)

yopito:
A rework might be need to support both qt 5.15 and qt < 5.15 ?

vitlav:
> A rework might be need to support both qt 5.15 and qt < 5.15 ?
this change is backward compatible (checked with 5.12.9)

killing:
We have added support for Qt 5.15 in the mainline. Closing this PR and thanks for the ideas!

PR #1300, opened 2020-06-03
support chose qtwebkit or qtwebengin

sun20121818:


moschlar:
I guess I'll close #1224 then

PR #1361, opened 2021-05-30
Added minor changes to get it to compile on Apple M1

tkreindler:
With the buildpath I'm outlining [here](https://forum.seafile.com/t/native-support-for-apple-silicon/13390/12) in order to compile for the Apple M1 you need to use the homebrew qt@5 formula instead of qt 5.6. In order to do that, very minor changes need to be done just to change updated dependencies.

These changes certainly don't need to be merged if you find another way to build for the M!, but I'm making this for visibility. 

The changes to seafile-applet.entitlements probably aren't necessary in the official buildpath but are for anyone who wants to self-sign their executable through xcode.


tkreind:
To add, I got the rest of my build directions from [here](https://download.seafile.com/published/seafile-manual/backup/build_seafile/osx.md).

PR #1389, opened 2021-12-02
Add "Toggle tray icon color" to macOS

makidoll:
Screen Shot 2021-12-01 at 20 16 09

The Seafile client's tray icon does not look good on Mac, so I've been researching how to fix the problem and here's what I've found:

- Looking at the source code, the client checks if the OS is in light or dark theme mode. This is wrong because the icon is either white or black depending on the user's background. Changing the theme doesn't change the icons color, so I removed the code checking for dark mode because it's obsolete. 

- According to the Apple guidelines (https://developer.apple.com/design/human-interface-guidelines/macos/extensions/menu-bar-extras), you need to use a template image of one color and set it as a mask, then the OS will automatically change its color when the background changes. Qt supports this with `QIcon::setIsMask(bool isMask)` which is great, but the Seafile client uses some color in the icon. If I enable it, it looks like this:
  Screen Shot 2021-12-01 at 20 24 11
  I thought, maybe I could redesign the icons so that it differentiates the colored bits using a transparent border from the main icon, but it would make the icon look a lot worse. It's a potential solution but I didn't want to do it.

- There's an app named Krisp which seems to do what we need, which looks like this:
  Screen Shot 2021-12-01 at 20 27 01
  I don't know how they did it (because they're not open source 😞) but after researching, I found that Xcode can make *imagesets* which contains multiple images that are light, dark, 2x, 3x, etc and it will use them intelligently. I made an imageset for one of the tray icons but there was no way I could get Qt to understand it, or pass it on directly to the OS. It might be possible to load the imageset through Objective C, but this probably means reimplementing the tray menu outside of Qt. Alternatively, it might be possible to just fetch the current tray menu from Objective C and update the icon from there, instead of through Qt. Either way, this seemed far to complicated for me and I didn't look into it further. It's unfortunate that Qt doesn't allow setting a light or dark variant in QIcon and then passing it on correctly to the tray.

Just this morning, I was playing with a program named Heroic Games Launcher which had a dark icon. But in the tray menu, there was a setting to toggle it between light or dark, and then I realized this is the easiest solution to this problem right now. So I went ahead and implemented it in the Seafile client and added this:

Screen Shot 2021-12-01 at 20 37 59

It works great! I added a setting to the settings manager named `trayIconDark` which is set to false by default. This is of course not the most elegant solution, but it's so much better than having a black icon.

I hope my code will do, but let me know if you'd like me to change something. Thank you so much for reading and considering my pr!

### DIFF
--- a/i18n/seafile_en.ts
+++ b/i18n/seafile_en.ts
@@ -2688,6 +2688,10 @@ File path contains invalid characters. It is not synced to this computer.</sourc
         <source>have some sync error</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Toggle tray icon color</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>SearchResultListView</name>
--- a/src/settings-mgr.cpp
+++ b/src/settings-mgr.cpp
@@ -30,6 +30,9 @@ namespace
 const char *kHideMainWindowWhenStarted = "hideMainWindowWhenStarted";
 const char *kHideDockIcon = "hideDockIcon";
 const char *kEnableSyncingWithExistingFolder = "syncingWithExistingFolder";
+#if defined(Q_OS_MACOS)
+const char *kTrayIconDark = "trayIconDark";
+#endif
 const char *kBehaviorGroup = "Behavior";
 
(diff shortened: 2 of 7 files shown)


killing:
This has been fixed in 9.0.5.

PR #1422, opened 2022-09-07
Fix app not moving in wayland by using the startSystemMove() method

### LINKED ISSUE #991: Title bar on Wayland is not working as expected
On Wayland session:
* Title bar can't move the window
* The minimize button is not working
* Window have no icon in taskbar
image

System:
* Kubuntu 17.04 with backports
* KDE Plasma: 5.10.5
* Qt: 5.7.1

alpemwarrior:
Fixes #991, where the app could not be moved under wayland because the protocol does not allow windows to move themselves. This method was introduced in qt 5.15 and [is currently recommended when using client-side decorations](https://www.qt.io/blog/custom-window-decorations) as it allows for native look and feel. 

I have only tested this under under wayland and xwayland but it should be a drop-in replacement in Windows, Linux and Macos. Testing would be appreciated nonetheless.

### DIFF
--- a/src/ui/cloud-view.cpp
+++ b/src/ui/cloud-view.cpp
@@ -289,22 +289,9 @@ void CloudView::onCloseBtnClicked()
 bool CloudView::eventFilter(QObject* obj, QEvent* event)
 {
     if (obj == mHeader) {
-        static QPoint oldPos;
         if (event->type() == QEvent::MouseButtonPress) {
-            QMouseEvent* ev = (QMouseEvent*)event;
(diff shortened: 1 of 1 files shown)


alpemwarrior:
Any feedback on this?

feliscat:
This does fix the problem for me in the latest KDE/Wayland under Arch.  It would be nice if it was merged.

whao:
Any update on this?

alpemwarrior:
I've rebased this on top of the current master (not that there were any conflicts).

PR #1426, opened 2022-10-10
Fix an error check in utils

rumtid:


killing:
Fixes https://github.com/haiwen/seafile-client/pull/1426

PR #1433, opened 2022-12-21
Notify users before bulk deleting files

rumtid:


src/message-poller.cpp
```diff
@@ -200,5 +204,24 @@ void MessagePoller::processNotification(const SyncNotification& notification)
...
+        QJsonDocument doc = QJsonDocument::fromJson(content.toUtf8(), &error);
+        if (doc.isNull()) {
+            qWarning() << "Failed to parse json:" << error.errorString();
+            return;
+        } else if (!doc.isObject()) {
+            qWarning() << "Malformed json string:" << content;
+            return;
+        }
+
+        QString msg = tr("Confirm to bulk delete files in library \"%1\" ?")
```
killing, line 219: Do you want to delete %1 files from library \"%2\" ?

需要打印准备删除多少个文件。

PR #1534, opened 2024-05-14
Fix the limit exceeded errors

rumtid:


src/filebrowser/auto-update-mgr.cpp
```diff
@@ -241,7 +241,9 @@ void AutoUpdateManager::onUpdateTaskFinished(bool success)
         } else if (task->httpErrorCode() == 441) {
             error_msg = tr("File does not exist");
         } else if (task->httpErrorCode() == 442) {
-            error_msg = tr("File count limit exceeded");
+            error_msg = tr("File size exceeds limit");
+        } else if (task->httpErrorCode() == 447) {
+            error_msg = tr("File quantity exceeds limit");
```
killing, line 246: Number of file exceeds limit

PR #1555, opened 2024-08-02
Unset LD_LIBRARY_PATH before openUrl

feiniks:


src/utils/utils.cpp
```diff
@@ -838,3 +838,18 @@ QString trimNULL(QString& s) {
     }
     return s;
 }
+
+bool openUrl(QUrl url) {
+#ifdef Q_OS_LINUX
+    // On linux openUrl will open the default browser via xgd-open, at this time LD_LIBRARY_PATH may affect xdg-open.
```
killing, line 844: On linux openUrl will open the default browser via xgd-open. If the client is packaged as AppImage format,  LD_LIBRARY_PATH is set to package location and may affect xdg-open.

PR #1558, opened 2024-08-13
Unsync repo that don't have permissions

feiniks:


src/repo-service.cpp
```diff
@@ -284,6 +284,19 @@ void RepoService::onRefreshSuccess(const std::vector<ServerRepo>& repos)
...
+        bool found = false;
+        const LocalRepo& repo = local_repos_[i];
+        for (size_t j = 0; j < server_repos_.size(); ++j) {
+            if (repo.id == server_repos_[j].id) {
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            seafApplet->rpcClient()->unsync(repo.id);
```
killing, line 297: 这里需要打印日志。

Issue #1568, opened 2024-10-14
Appimage misses an icon.


deragon:
Appimage for Seafile client 9.0.8 Rev 1 does not contain an icon that can be used by Gnome (and probably other desktop environment (DE) such as KDE Plasma).  The DE must then use a generic icon as show below.

The  screenshot has been made on Gnome 46 running under Wayland in Ubuntu 24.04 LTS.

The screenshot also shows other application missing an icon, thus the reason why three applications are shown with the same icon.

![image](https://github.com/user-attachments/assets/9a29a616-1b92-423a-964c-6340a5c25362)

killing:
Please refer to https://docs.appimage.org/user-guide/run-appimages.html#ref-desktop-integration for desktop integration to display AppImage's application icons.

PR #1570, opened 2024-10-30
Fix wayland window associations by renaming .desktop file

ReillyBrogan:
For Wayland window managers to correctly associate open windows with the corresponding .desktop file the name of the `.desktop` file sans `.desktop` suffix needs to match the window appId.

By default Qt sets the appId to the name of the binary, but `QCoreApplication::setOrganizationDomain("seafile.com");` in `src/main.cpp` causes it to use the domain to create a fully qualified appId (`com.seafile.seafile-applet`). This does not match the desktop file `seafile.desktop`, which causes open windows to not be correctly associated with the launcher. While it's possible to use `QGuiApplication::setDesktopFileName()` to override the appId it's more consistent with FreeDesktop specs to use the fully qualified appId instead.

To see the difference this makes it is sufficient to start seafile-applet in a Plasma Wayland session (which has easily accessible window property introspection tools) with and without `com.seafile.seafile-applet.desktop` being the name of the `.desktop` file. Without the window will show the generic Wayland icon in the overview, with it will show the seafile icon shipped in the package.

Resolves haiwen/seafile-client#1568 (probably)


killing:
Thanks! We'll have a look.

PR #1613, opened 2025-04-08
CMakeLists.txt: require WebEngine{Core,Widgets} only if BUILD_SHIBBOL…

clan:
…ETH_SUPPORT

which is handled in 'IF (BUILD_SHIBBOLETH_SUPPORT)'


rumtid:
Thank you for your pull request. It appears that the variables LIBEVENT_INCLUDE_DIRS, LIBEVENT_LIBRARY_DIRS, and LIBEVENT_LIBRARIES can also be removed from CMakeLists.txt.

PR #1638, opened 2025-08-03
Add Arabic translation for the client

GhaziTriki:
Added ar_SA i18n. A review might be needed. I tested the UI, it switches to RTL by default but some layout improvements might be required.


GhaziTriki:
I can see now after being accepted in Transifex that Arabic locale is "ar". Should I update my PR?

PR #1644, opened 2025-08-22
Rename ar_SA to ar

rumtid:


rumtid:
Hi, @GhaziTriki.

Under normal circumstances, we first complete the translation in Transifex, and then add the new language in the code.

We noticed that you added ar_SA in the code. Is there any difference between the ar_SA and ar translations? If not, we’d prefer to simplify things by keeping just one language (ar).

PR #1670, opened 2026-02-17
Add missing include for Qt 6.9+

### LINKED ISSUE #1664: Support Qt 6.9+
When trying to build against Qt 6.9+, I get the following error:
```
/run/build/seafile-client/src/message-poller.cpp: In member function ‘void MessagePoller::processNotification(const SyncNotification&)’:
/run/build/seafile-client/src/message-poller.cpp:247:30: error: variable ‘QRegularExpression re’ has initializer but incomplete type
  247 |         QRegularExpression re("Deleted \"(.+)\" and (.+) more files.");
      |                              ^
/run/build/seafile-client/src/message-poller.cpp:248:49: error: invalid use of incomplete type ‘const class QJsonValue’
  248 |         auto match = re.match(doc["delete_files"].toString().trimmed());
      |                                                 ^
In file included from /usr/include/QtCore/qobject.h:18,
                 from /usr/include/QtCore/qabstracteventdispatcher.h:7,
                 from /usr/include/QtCore/qbasictimer.h:8,
                 from /usr/include/QtCore/qtimer.h:11,
                 from /usr/include/QtCore/QTimer:1,
                 from /run/build/seafile-client/src/message-poller.cpp:1:
/usr/include/QtCore/qmetatype.h:1544:1: note: forward declaration of ‘class QJsonValue’
 1544 | QT_FOR_EACH_STATIC_CORE_CLASS(QT_FORWARD_DECLARE_STATIC_TYPES_ITER)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This can probably be fixed by adding a missing include.

darkdragon-001:
Fixes #1664

### DIFF
--- a/src/message-poller.cpp
+++ b/src/message-poller.cpp
@@ -1,6 +1,7 @@
 #include <QTimer>
 #include <QDateTime>
 #include <QJsonDocument>
+#include <QJsonValue>
 
 #include "utils/utils.h"
 #include "utils/translate-commit-desc.h"


darkdragon-001:
@killing Adding direct includes for all used symbols instead of relying on transitive includes is good coding practice anyways. So this should not break compatibility with older Qt versions.

